Issue #141, opened 2020-10-29
Capture Wall


seanwu1105:
Implement the UI for Capture/Post Capture Wall.

![Screenshot from 2020-10-29 12-15-18](https://user-images.githubusercontent.com/14951000/97524789-6ef74680-19e0-11eb-9cb4-f001077f82f9.png)

![Screenshot from 2020-10-29 12-26-34](https://user-images.githubusercontent.com/14951000/97525551-08732800-19e2-11eb-8df5-c3d77f1cd6e2.png)

scott-dt:
請見 https://docs.google.com/presentation/d/1lnXs-dMzKT7IiAce4GdE9KLxohg2SFAWdlb-x0yaxtE/edit?usp=sharing

scott-dt:
v0.7.2 修正完成

PR #163, opened 2020-11-04
Feature capture wall

seanwu1105:
See #141.


src/theme/variables.scss
```diff
@@ -1,3 +1,39 @@
...
+      accent: $app-accent,
+      warn: $app-warn,
+    ),
+  )
+);
+
+// Include theme styles for core and each component used in your app.
+// Alternatively, you can import and @include the theme mixins for each component
+// that you are using.
+@include angular-material-theme($app-theme);
```
shc261392, line 35: For the Codacy stylelint error, maybe it would be better to create a `.styleintrc.json` and use the SASS guideline configs. (Provided that the Codacy Code Analysis reads project stylelint configs)
https://github.com/bjankord/stylelint-config-sass-guidelines

shc261392 (commented):
Please check the comments

src/theme/variables.scss
seanwu1105: I think adding the SASS guideline configs would increase the complexity of the project for now. We can simply ignore the error messages in the dashboard of Codacy. In the future, we should introduce prettier to format all codes in all languages instead of adding the style config separately.

shc261392 (approved):

PR #204, opened 2020-11-16
Fix #202, #201 and #198

### LINKED ISSUE #202: Proof Detail => Capture Details（影像資訊）
## Description
相片點入後 title 英文  Proof Detail 修正為 Capture Details, 中文 拍攝資訊 修正為 影像資訊

seanwu1105:
See #202, #201 and #198.

The fix for #198 is a workaround.

```
Show sender on app-post-capture-card. #198
    
    This is implemented with a BAD architecture. The PostCapture wall only display contents from the server,
    and the app does NOT save any cache of PostCaptures. The performance will deteriorate for the huge
    ovehead from the Internet. This architecture should be redesign with the new (reliable) storage
    implementation and the new data model for the exchange between server and app.
```


shc261392 (approved):
Looks good to me. Not tested on real device.

Issue #209, opened 2020-11-16
User name should be mandatory 


scott-dt:
## Description

修改 sign-up UI 讓 user 輸入 username

scott-dt:
如果沒有填寫 username 註冊按鈕則不會亮起 無法進行註冊
![Screenshot_20201127-154216_Capture](https://user-images.githubusercontent.com/62582260/100423649-48075f80-30c7-11eb-8b78-5b441b11be0d.jpg)
![Screenshot_20201127-154229_Capture](https://user-images.githubusercontent.com/62582260/100423651-49388c80-30c7-11eb-94ca-2a0118d41804.jpg)

PR #224, opened 2020-11-17
Required User Name on Sign-Up Page

seanwu1105:
See #209.


shc261392 (changes requested):
Username could be called `username` or `user name` (camel case `userName`), but it's better to stick to one form.
The API request body requires the field to be `username`. The naming of this property should be `username` too if there's no reason against it.

shc261392 (approved):
Looks great to me

PR #227, opened 2020-11-17
Improve Stability and Reliability of Data Persistence

### LINKED ISSUE #150: Avoid Using JSON.stringify When Hash Objects in Storage Utils
Use sha256 is not a good idea as `sha256$()` uses `JSON.stringify()` under the hood. Thus, the order of the tuple (properties) will result in different hash if tuple is an array (object).

Maybe we need to use SQLite to save files.

seanwu1105:
Implement stable and reliable data persistence solution with Capacitor `Filesystem` plugin.

Use the plugin to create a JSON file for each table in database. The JSON contains a single big array to store tuples. Also, use an `async-mutex` to insure each insertion and deletion is atomic. Write a fully covered unit tests to make sure the reliability.

Fix #150.


shc261392 (approved):
The improvement seems solid.

PR #252, opened 2020-11-25
Refactor Proof Interface

seanwu1105:
Due to the limitation of Capacitor and Ionic framework, we cannot and should not use the same architecture from Starling Capture, which is based on Android framework. Thus, I have rewritten most parts regarding the data persistence. This includes:

- Merge `InformationRepository`, `SignatureRepository` and `SerializationService` into `ProofRepository` and provide the same functionality.
- Refactor `CollectorService` and the corresponding `InformationProvider`s and `SignatureProvider`s to accept newly merged `Proof` instance.
- Refactor `PublisherAlert` and the corresponding publisher `NumbersStoragePublisher` and `NumbersStorageApi` to accept newly merged `Proof` instance.
- Write unit tests to fully cover the refactored components, and thus the coverage has reached 70%.
- Implement an adapter (`old-proof-adapter`) to provide the connection between Numbers Storage Back-end.


shc261392 (approved):
LGTM

PR #269, opened 2020-11-27
Refactor PreferenceManager and Preferences

seanwu1105:
Refactor `PreferenceManager` and `Preferences` with fully covered tests. This new implementation meets the same quality to the one of Data Persistence refactoring (#227) by using `async-mutex` to make sure each `set` methods is atomic.


shc261392 (approved):
Looks good to me

PR #321, opened 2020-12-09
Implement the not-uploaded overlay and re-upload click event for Captures

shc261392:
## Enhancements

- Change data flow to allow undefined asset so that upload failure doesn't prevent the view from updating.

- Display semi-transparent white overlay and error icon when upload fails. Clicking the overlay triggers re-uploading.

- Display uploading spinner when re-uploading.

## Limitations

- The user can only re-upload 1 photo at a time. During the upload, tapping the other upload-error photos does nothing.

- When uploading the photo in the normal flow, the user would see the upload error first (after the proof is added to the repository but before the asset is added to the repository)

It seems to me that trying to remove these limitations without adding a persistent upload state for Proof would make the component data flow very dirty. Feature-wise I believe the current limitation is acceptable so maybe we could improve it later, but if there's any quick way to improve the implementation or remove the limitations I'd be glad to hear it and make change accordingly.

### DIFF
--- a/src/app/pages/home/home.page.scss
+++ b/src/app/pages/home/home.page.scss
@@ -44,6 +44,23 @@ mat-toolbar {
     border-radius: 4px;
     background-position: center center;
     background-size: cover;
+
+    .image-tile-overlay {
+      width: 100%;
+      height: 100%;
+      background-color: white;
+      opacity: 0.7;
+    }
+
+    .overlay-icon {
+      opacity: 1;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      -ms-transform: translate(-50%, -50%);
+      text-align: center;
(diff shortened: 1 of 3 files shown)


src/app/pages/home/home.page.scss
```diff
@@ -44,6 +44,23 @@ mat-toolbar {
...
+      opacity: 0.7;
+    }
+
+    .overlay-icon {
+      opacity: 1;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      -ms-transform: translate(-50%, -50%);
```
seanwu1105, line 61: Unnecessary rule.

seanwu1105 (changes requested):

seanwu1105 (approved):

PR #340, opened 2020-12-13
Fix dia backend auth migrate

seanwu1105:
Completely migrate the `DiaBackendAuthService` from old `NumbersStoragePublisher` class. This migration has been tested on web and Android devices with 0.7.5 to 0.11.4 update. The previous failure is caused by the auth `token` contains `token ` string in the old version but not in the latest one.

### DIFF
--- a/src/app/services/dia-backend/auth/dia-backend-auth.service.ts
+++ b/src/app/services/dia-backend/auth/dia-backend-auth.service.ts
@@ -30,7 +30,10 @@ export class DiaBackendAuthService {
       key: 'numbersStoragePublisher_authToken',
     });
     if (oldToken.value) {
-      this.setToken(oldToken.value.split(' ')[1]);
+      const splitted = oldToken.value.split(' ');
+      if (splitted[0] === 'token' && splitted[1]) {
+        this.setToken(splitted[1]);
+      }
     }
 
     const oldUsername = await Storage.get({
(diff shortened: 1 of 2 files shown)


src/app/services/dia-backend/auth/dia-backend-auth.service.ts
```diff
@@ -30,7 +30,21 @@ export class DiaBackendAuthService {
       key: 'numbersStoragePublisher_authToken',
     });
     if (oldToken.value) {
-      this.setToken(oldToken.value);
+      this.setToken(oldToken.value.split(' ')[1]);
```
shc261392, line 33: The token will be set to `undefined` if the original token doesn't contain a space and prevents redirection to the login page as well as communication with dia-backend. Though it seems unlikely to happen, we should either make sure all previously released version has an expected token value format, or make the login check fail for `undefined` token value (if the token is malformed it's not gonna work anyway)

https://github.com/numbersprotocol/capture-lite/blob/b156980f4a0ac3a5355f82ad362d2522d03e25b9/src/app/services/dia-backend/auth/dia-backend-auth.service.ts#L138

seanwu1105:
@shc261392 I have updated the logic for both `hasLoggedIn` method

https://github.com/numbersprotocol/capture-lite/blob/b156980f4a0ac3a5355f82ad362d2522d03e25b9/src/app/services/dia-backend/auth/dia-backend-auth.service.ts#L138

and `migrate` method to avoid accepting the malformed token and setting the new token to `undefined`.

shc261392 (approved):

PR #341, opened 2020-12-13
Fix drop pictures during capturing

seanwu1105:
See #329. This PR uses a workaround to fix an upstream bug: https://github.com/ionic-team/capacitor/issues/3062. The fix has been tested on Exodus 1 and Exodus 1s. However, no iOS devices has been tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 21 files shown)


shc261392 (changes requested):
On iPhone 7 Plus, iOS 13.6, tested indoor (connected to WiFi), it always takes about 12~15 seconds to get the current position. If using the default timeout (10s), the location collection will always result in timeout error. I'd suggest changing the timeout duration to 20s.

shc261392 (approved):
Tested on iPhone 7 Plus by taking 10 Captures consecutively. No Capture is dropped, so the workaround is valid. However, only 2 out of 10 successfully collected location info, and only 8 out of 10 is uploaded successfully.

The behavior is acceptable in this release.

PR #343, opened 2020-12-13
Change the background-size to cover for capture thumbs.

seanwu1105:
See #331. The repeated thumbnail for captures is fixed by setting the `background-size` CSS property to `cover`. The fix has been tested on web and Exodus 1s.


shc261392 (approved):
Tested on iPhone 7 Plus, works as expected.

PR #344, opened 2020-12-13
Show errors on snackbar and notification

seanwu1105:
See #239. This feature has been tested on web and Exodus 1s.


shc261392 (changes requested):
From what I know showing error message on both the snackbar and local notification doesn't seem to match the description of feature request `Capture 的 Information 以及 error message 修改為直接在 APP 中跳出視窗`

I'd suggest getting confirmation to make sure this is the expected requested feature (and update the issue description if needed).

seanwu1105:
Close due to the discussion.

> Tammy: 因為 collecting proof & error 是兩個不同性質的 notification
> collecting proof => passive notification => user don't need to do anything
> error => real notification => user needs to take action

PR #345, opened 2020-12-13
Add extension to the image stored in ImageStore.

seanwu1105:
Add file extension to the images stored in `ImageStore` in order to successfully share files with Capacitor `Share` plugin. We do not need to migrate the data if we merge this PR in next release.


src/app/services/image-store/image-store.service.ts
```diff
@@ -139,6 +148,35 @@ export class ImageStore {
     return thumbnails.find(thumb => thumb.imageIndex === index);
   }
 
+  private async getExtension(index: string) {
```
shc261392, line 151: The function name collides with `getExtension` from `mime-type.ts` so it's a bit confusing at the first glance because they do very different things.

Do you think `toExtension` would be a better name for the function in `mime-type.ts`? (Since another function is already called `fromExtension`)

shc261392 (changes requested):

shc261392 (approved):

PR #346, opened 2020-12-13
Feature share watermarked postcapture

seanwu1105:
See #311. This feature has been tested on Exodus 1s.


shc261392 (approved):
Everything looks good and tested on iPhone 7 Plus (iOS 13.6) successfully.

PR #347, opened 2020-12-14
Fix iOS DIA backend delete error

shc261392:
Append slash in delete asset API endpoint to prevent iOS webview dropping auth header when HTTP redirection.


seanwu1105:
Put a note here: This PR can address the issue #330.

seanwu1105 (approved):

PR #361, opened 2020-12-16
Wait until All Thumbnails Loaded

seanwu1105:
Wait until all thumbnails is loaded before display on screen, and sort captures descendingly in groups.


shc261392 (approved):
The fix looks valid.

PR #373, opened 2020-12-16
Feature use server side notification for push

shc261392:
- Update language settings to DIA backend.

- Replace local notification for push notification data with built-in notification of push notification.

- Workaround: Force change detection when the inbox emits value to update the badge when returning to the App foreground by clicking the notification in background.


src/app/services/dia-backend/auth/dia-backend-auth.service.ts
```diff
@@ -109,6 +117,24 @@ export class DiaBackendAuthService {
...
+    return combineLatest([
+      this.getAuthHeaders$(),
+      this.languageService.currentLanguageKey$,
+    ]).pipe(
+      distinctUntilChanged(isEqual),
+      concatMap(([headers, language]) =>
+        this.httpClient.patch(
+          `${BASE_URL}/auth/users/me/`,
+          {
+            language,
```
seanwu1105, line 130: This line can have a better format.

seanwu1105 (changes requested):
I have tested on Android Exodus 1s. However, I cannot receive any push notification whenever the app is on foreground or background.

shc261392:
The updated code could work for both iOS and Android, foreground and background. Tested on iPhone 7 Plus (iOS 13.6) and Nokia 6.1 (Android 10).

Known issue: the notification icon of Android when receiving the notification in background is incorrect (white block icon).

seanwu1105 (approved):

PR #409, opened 2020-12-21
fix ios Empty Inbox position

CS6:
image


src/app/pages/home/inbox/inbox.page.scss
```diff
@@ -11,6 +11,10 @@
     justify-content: center;
   }
 
+  .empty-inbox {
+    padding-top: 80px;
```
seanwu1105, line 15: Why not merge this rule to `span.empty-inbox` selector?

seanwu1105 (changes requested):

PR #410, opened 2020-12-21
fix font &  position

CS6:
image


src/app/pages/tutorial/tutorial.page.scss
```diff
@@ -21,6 +21,10 @@ ion-content {
   }
 }
 
+.ios text {
```
seanwu1105, line 24: This rule only apply to iOS devices. Is this what we want?

seanwu1105 (changes requested):

seanwu1105 (approved):

PR #441, opened 2021-01-06
WORKAROUND: Fix Unable to Send PostCaptures

seanwu1105:
Introduce yet another workaround for clear the unsynced catch regarding PostCapture deletion. Also, use `shareReplay({bufferSize: 1, refCount: true})` to replace `share()` to avoid the shared `Observable` from not emitting on subscribe.


shc261392 (approved):
Looks like a viable workaround, no issue spotted.

PR #443, opened 2021-01-11
Regroup project structure.

seanwu1105:
To isolate the capture page and its dependent services, regroup the project structure. No features or functionality changed.

- Move shared services and utilities to `./app/shared/`
- Move pages to `./app/features/`
- Update `set-secret.js` to TypeScript


shc261392 (commented):
Type assertion is added for two tests but not mentioned. It would be better to mention every different kind of changes since the pull request modifies a lot of files.

If pages and only pages go to `features` folder, it is identical to renaming, so what is the purpose of this regrouping? Is it just preparing for further project structure changes?

seanwu1105:
> Type assertion is added for two tests but not mentioned. It would be better to mention every different kind of changes since the pull request modifies a lot of files.

@shc261392 Okay, I will mention all details in the future PR.

> If pages and only pages go to `features` folder, it is identical to renaming, so what is the purpose of this regrouping? Is it just preparing for further project structure changes?

The currently `pages` folder is misleading as it contains some `Component`s. Not every files within `pages` folder is technically an Ionic `Page`.

shc261392 (approved):
That answered my question. Btw It's ok not mentioning details if the PR is small (I mentioned it because this PR modifies 185 files and it's easy to overlook changes that are not so obvious)

PR #451, opened 2021-01-13
Rebuild Capture - Step D2 (Non-blocking UI Upload)

shc261392:
Main feature:
- Add upload service to handle upload task queue and pause/resume
- Add upload-bar component
- Display diaBackendAssetId on capture-details page

Things done for convenience in order to implement the feature (probably not best solution):
- Add diaBackendAssetId field to Proof
- Small adjustment to DiaBackendAssetRepository


src/app/shared/services/upload/upload.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    const taskDebounceTime = 50;
+    return combineLatest([
+      this.proofRepository.getAll$().pipe(debounceTime(taskDebounceTime)),
+      this.isPaused$,
+    ]).pipe(
+      tap(([proofs, isPaused]) => {
+        const tasks = proofs.filter(
+          proof => !proof.diaBackendAssetId && !proof.willCollectTruth
+        );
+        this._currentUploadingCount$.next(tasks.length);
```
seanwu1105, line 81: I think we should update the `this._currentUploadingCount` in [`this.uploadTaskQueue`](https://github.com/numbersprotocol/capture-lite/pull/451/files#diff-566b2bc9075a195ca816285e30a40af05260703c2aaf60a93dc82a06fc67cbe9R107).

Also, why do we need `this._currentUploadingCount$`? I think generating `this.currentUploadingCount$` from `this.taskQueue$` could be a good idea?

src/app/shared/services/repositories/proof/proof.ts
```diff
@@ -7,6 +7,7 @@ import { ImageStore } from '../../image-store/image-store.service';
 
 export class Proof {
   willCollectTruth = false;
+  diaBackendAssetId = '';
```
seanwu1105, line 10: I believe using `undefined` instead of `''` empty string could be safer as TypeScript prefers [Nullish Coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing), which won't work if we use empty string as default value.

Also, `IndexedProofView` use the same nullable property.

src/app/shared/services/dia-backend/asset/dia-backend-asset-repository.service.ts
```diff
@@ -68,15 +82,7 @@ export class DiaBackendAssetRepository {
...
-  async add(proof: Proof) {
-    return this.notificationService.notifyOnGoing(
-      this.createAsset$(proof),
-      this.translocoService.translate('registeringProof'),
-      this.translocoService.translate('message.registeringProof')
-    );
-  }
-
-  private createAsset$(proof: Proof) {
+  createAsset$(proof: Proof) {
```
seanwu1105, line 85: (Only suggestion) I think we should rename this method to `add$(proof: Proof)` for the consistency of the method names of a "Repository".

src/app/features/home/capture-tab/upload-bar/upload-bar.component.html
```diff
@@ -0,0 +1,44 @@
...
+  *ngIf="{
+    currentUploadingCount: currentUploadingCount$ | async,
+    isPaused: isPaused$ | async,
+    isPausedByFailure: isPausedByFailure$ | async
+  } as uploadState"
+>
+  <mat-toolbar color="background" class="thin-upload-bar">
+    <ng-container *ngIf="uploadState.currentUploadingCount > 0; else completed">
+      <span class="upload-text" *ngIf="!uploadState.isPaused"
+        >Registering {{ uploadState.currentUploadingCount }} photos on
```
seanwu1105, line 11: Need i18n after demo.

seanwu1105 (changes requested):

src/app/shared/services/upload/upload.service.ts
shc261392: The initial design is different from the current implementation, so it's a good suggestion I overlooked.

src/app/shared/services/dia-backend/asset/dia-backend-asset-repository.service.ts
shc261392: It makes sense since it is still Repository now. I will make changes according to the suggestion.

src/app/shared/services/upload/upload.service.ts
shc261392: The `currentUploadingCount$` holds the task count that can be exposed and displayed on UI and it should not be cleared when pausing, while the taskQueue$ will be cleared when pausing. The component also uses `currentUploadingCount$` to determine if all tasks are completed, so filtering out zero value cannot be a solution.

I haven't come up with a decent solution yet, but I think this could be improved later.

PR #452, opened 2021-01-13
Rebuild Capture - Step D1 (Stateful Capture Items)

seanwu1105:
- Add mutex lock when updating database table.
- Navigate up after deleting capture item in capture-details page.
- Remove notificatinos when collecting and uploading proofs.
- Implement stateful `CaptureItem` and `CaptureItemComponent`.
- Display an empty small image to avoid border of `img` without `src`.
- Remove spinners in inbox and transactions pages.


src/app/features/home/capture-tab/capture-tab.component.ts
```diff
@@ -39,4 +33,17 @@ export class CaptureTabComponent {
...
+  trackCaptureGroupByDate(
+    _: number,
+    item: { key: string; value: CaptureItem[] }
+  ) {
+    return item.key;
+  }
+
+  // tslint:disable-next-line: prefer-function-over-method
+  trackCaptureItem(_: number, item: CaptureItem) {
+    return `${item.oldProofHash}_${item.proof?.willCollectTruth}`;
```
shc261392, line 47: It seems track by both `oldProofHash` and `willCollectTruth` makes Angular to re-render the CaptureItem when truth is collected, causing obvious flickering. I'm not sure about the necessity of combining them as tracking index here.

src/app/features/home/capture-tab/capture-item/capture-item.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+@UntilDestroy({ checkProperties: true })
+@Component({
+  selector: 'app-capture-item',
+  templateUrl: './capture-item.component.html',
+  styleUrls: ['./capture-item.component.scss'],
+})
+export class CaptureItemComponent {
+  @Input()
+  set item(value: CaptureItem) {
```
shc261392, line 17: What is the different between this approach (creating a BehaviorSubject in the component and pass the object from parent component through Input property) and directly use the Input property?

The CaptureItem from parent component is created from a subscribed observable, wouldn't it be propagated and updated automatically by Angular, or does it have something to do with avoiding additional re-rendering?

shc261392 (approved):

PR #453, opened 2021-01-14
Use in-memory set to record proofs status.

seanwu1105:
- Remove the error-prone repository initialization to avoid multiple DOM refresh and `ImageStore.read`.
- Use an in-memory `Set` to store the collecting state of proofs.


shc261392 (approved):
Looks great. `ImageStore.read()` is once called once per CaptureItem during initialization.

PR #480, opened 2021-01-25
Show prefetching/tutorial pages only on first time login.

seanwu1105:
Migrate the `IS_ONBOARDING` preference in `OnboardingService` to `HAS_SHOWN_TUTORIAL` and `HAS_PREFETCHED_DIA_BACKEND_ASSETS` preferences to avoid showing these pages again if the user has logged in before.

### DIFF
--- a/src/app/shared/services/migration/migration.service.ts
+++ b/src/app/shared/services/migration/migration.service.ts
@@ -1,5 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Plugins } from '@capacitor/core';
+import { BehaviorSubject } from 'rxjs';
+import { distinctUntilChanged } from 'rxjs/operators';
 import {
   DiaBackendAsset,
   DiaBackendAssetRepository,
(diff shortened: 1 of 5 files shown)


src/app/shared/services/migration/migration.service.ts
```diff
@@ -74,6 +86,16 @@ export class MigrationService {
     }
     return ret;
   }
+
+  private async updateOnboardingServicePreferences() {
+    const preferences = this.preferenceManager.getPreferences(
+      OnboardingService.name
+    );
+    if (!(await preferences.getBoolean('IS_ONBOARDING'))) {
```
shc261392, line 94: If `IS_ONBOARDING` (an preference which is only truthy if the user finished onboarding process in old versions) is falsy => set the new onboarding preferences truthy
doesn't seem right

shc261392 (changes requested):

shc261392 (approved):

PR #484, opened 2021-01-26
If writing mode is not REPLACE or IGNORE, throw an exception.

bafu:
If writing mode is not REPLACE or IGNORE, throw an exception.

Signed-off-by: Bofu Chen (bafu) <bofu@numbersprotocol.io>


seanwu1105:
Superseded by #489.

PR #487, opened 2021-01-26
Fix onboarding flow

seanwu1105:
- Remove `isNewLogin` condition wrapping `redirectOnboarding()` on home page.
- Set `isNewLogin` to `false` before checking if to prefetch dialog.
- Set `hasPrefetchedDiaBackendAssets` to `true` after migration.


src/app/features/home/home.page.ts
```diff
@@ -49,15 +49,8 @@ export class HomePage {
...
       .pipe(
-        switchTap(isNewLogin =>
-          defer(() => {
-            if (isNewLogin) {
-              return this.onboardingRedirect();
-            }
-            return VOID$;
-          })
-        ),
         concatMap(isNewLogin => this.migrationService.migrate$(isNewLogin)),
```
shc261392, line 52: Should catch errors for migrate$. If migrate$ fails, `onboardingRedirect()` still needs to go on.

shc261392 (changes requested):

shc261392 (approved):

PR #490, opened 2021-01-27
Fix migrating set has prefetch corner case

shc261392:
Fix the following conditions:
- If an error happens during migrating, the migration dialog will not be closed.
- If an error happens during migrating, the`HAS_PREFETCHED_DIA_BACKEND_ASSETS` preference will not be set to true while it should.
- If network is not connected (assuming that user is using the App offline), the migrating dialog will prompt and disappear everytime the user navigates to the home page.
- If an error happens during pre-fetching, the pre-fetching dialog will not be closed.

### DIFF
--- a/src/app/features/home/onboarding/prefetching-dialog/prefetching-dialog.component.ts
+++ b/src/app/features/home/onboarding/prefetching-dialog/prefetching-dialog.component.ts
@@ -20,9 +20,15 @@ export class PrefetchingDialogComponent {
   }
 
   private async prefetch() {
-    await this.diaBackendAssetPrefetchingService.prefetch(
-      (currentCount, totalCount) => (this.progress = currentCount / totalCount)
-    );
+    await this.diaBackendAssetPrefetchingService
+      .prefetch(
+        (currentCount, totalCount) =>
+          (this.progress = currentCount / totalCount)
+      )
+      .catch(err => {
+        this.dialogRef.close();
+        throw err;
+      });
     await this.onboardingService.setHasPrefetchedDiaBackendAssets(true);
     this.dialogRef.close();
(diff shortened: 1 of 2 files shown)


src/app/features/home/onboarding/prefetching-dialog/prefetching-dialog.component.ts
```diff
@@ -20,9 +20,15 @@ export class PrefetchingDialogComponent {
...
-    await this.diaBackendAssetPrefetchingService.prefetch(
-      (currentCount, totalCount) => (this.progress = currentCount / totalCount)
-    );
+    await this.diaBackendAssetPrefetchingService
+      .prefetch(
+        (currentCount, totalCount) =>
+          (this.progress = currentCount / totalCount)
+      )
+      .catch(err => {
+        this.dialogRef.close();
```
seanwu1105, line 29: Suggestion: Maybe we can use `try {} finally {}` to close the `dialogRef`?

src/app/shared/services/migration/migration.service.ts
```diff
@@ -20,47 +21,59 @@ const { Device } = Plugins;
...
+      throw new Error('No network connection, aborting migration.');
+    }
     const dialogRef = this.dialog.open(MigratingDialogComponent, {
       disableClose: true,
       data: { progress: 0 },
     });
 
-    await this.to0_15_0();
+    await this.to0_15_0().catch(err => {
+      dialogRef.close();
```
seanwu1105, line 74: Suggestion: Maybe we can use `try {} finally {}` to close the `dialogRef`?

seanwu1105 (approved):

PR #494, opened 2021-01-28
Revert: fix migration set has prefetch corner case

seanwu1105:
Revert the merge `fix-migration-set-has-prefetch-corner-case` due to the logic error reported by @scott-dt on qa-pm Slack channel.

> @scott-dt: 0.15.7 測試 新安裝使用舊帳號 restore 完成後 點選圖片後 又詢問一次 是否restore 如影片流程 測試三次都是一樣狀態

This revert has been tested on web and Exodus 1.


shc261392:
The root cause is that the account has assets with unsupported data format created by using API directly.
Currently, we do not support that kind of data.

PR #496, opened 2021-01-28
Show error dialog on prefetch error.

seanwu1105:


src/app/features/home/onboarding/prefetching-dialog/prefetching-dialog.component.ts
```diff
@@ -26,6 +30,12 @@ export class PrefetchingDialogComponent {
           (this.progress = currentCount / totalCount)
       );
       await this.onboardingService.setHasPrefetchedDiaBackendAssets(true);
+    } catch {
+      this.dialogRef.close();
+      await this.confirmAlert.present({
```
shc261392, line 35: Showing the confirm alert with two buttons (Cancel and OK) might confused the user. They don't have any difference in this case, but the user will be uncertain of that.

shc261392 (changes requested):

shc261392 (approved):

PR #499, opened 2021-01-29
Implement capture sharing.

seanwu1105:


src/app/features/home/capture-tab/capture-details/options-menu/options-menu.component.html
```diff
@@ -1,4 +1,7 @@
 <mat-nav-list *transloco="let t">
+  <a mat-list-item (click)="openLink(options.Share)">
```
shc261392, line 2: It might be a good idea to hide the share option if `diaBackendAssetID` is `undefined`, so the user won't be able to access the share button if the Capture can't be shared.

shc261392 (changes requested):

shc261392 (approved):

PR #509, opened 2021-02-01
Show capture status icons on capture-tab.

seanwu1105:
- Show media-id icon if the capture has been registered.
- Show location icon if the capture has geolocation information.
- Avoid clipping full image on capture-details page.


shc261392 (approved):
Looks good to me

PR #510, opened 2021-02-01
Remove view more information on PostCaptureCard.

seanwu1105:
I decide to remove the related codes completely instead of hiding them as the version control always remembers the codes, and thus could be easily referenced in the future if needed.


shc261392 (changes requested):
截圖 2021-02-01 下午12 40 33
The original caption has a 14px top-margin so that the caption text won't look too closed to the image. The margin is gone now and I feel that adding it back will make it look better and also retains the original style as much as possible.

(The style difference can be observed on both web and iOS)

shc261392 (approved):

PR #523, opened 2021-02-02
Remove email confirmation page after signup.

seanwu1105:
- User will be redirected to login page upon finishing the signup process.
- The redirected login page will automatically login with the new account just signed up.


shc261392 (approved):
Looks good to me

PR #525, opened 2021-02-02
Adjust options on the menu of capture-details.

seanwu1105:
- Add: Share moment on options menu
- Add: Send moment on options menu
- Add: Delete moment on options menu
- Add: View blockchain certificate on options menu
- Remove: view blockchain certificate on media ID


shc261392 (approved):
Looks exactly the same as the requirement.

PR #544, opened 2021-02-18
iOS CD

shc261392:


.github/workflows/pre-release.yml
```diff
@@ -70,6 +70,92 @@ jobs:
...
+        if: steps.version_check.outputs.changed == 'true'
+        with:
+          app-path: build/App.ipa
+          issuer-id: ${{ secrets.APPSTORE_ISSUER_ID }}
+          api-key-id: ${{ secrets.APPSTORE_KEY_ID }}
+          api-private-key: ${{ secrets.APPSTORE_PRIVATE_KEY }}
+
+      - name: Upload IPA to artifacts
+        if: steps.version_check.outputs.changed == 'true'
+        uses: actions/upload-artifact@v1
```
seanwu1105, line 147: I think we should not upload large files, such as IPA or Xcode archive, to artifacts due to the storage space limitation. We could upload the files to Google drive if needed. The related secret keys and configuration has been already set. You can check the `upload-google-drive` for the deteiled action config.

seanwu1105 (changes requested):
1. I do not think storing large files in the artifacts is a good idea. 
1. Maybe we can add `ios` job to the `build.yml` action for each push.

After the first pre-release has been succesfully deployed, we can add release URL to [the Slack message](https://github.com/numbersprotocol/capture-lite/pull/544/files#diff-8f0c4d810b263ab478547020c847ee0297cc98c11c67fc2209ce604e52719607R290).

seanwu1105 (approved):

PR #548, opened 2021-02-20
Open Google maps with external URL on clicking location info.

seanwu1105:
The maps will be opened with Capacitor `Browser` plugin. Thus, the map would be the website of Google map instead of Google map app.

This PR has been tested on Brave browser and Exodus 1.

### DIFF
--- a/src/app/features/home/capture-tab/capture-details/capture-details.page.ts
+++ b/src/app/features/home/capture-tab/capture-details/capture-details.page.ts
@@ -166,6 +166,26 @@ export class CaptureDetailsPage {
       .subscribe();
   }
 
+  openMap() {
+    return this.proof$
+      .pipe(
+        concatMap(proof =>
+          iif(
+            () =>
+              proof.geolocationLatitude !== undefined &&
+              proof.geolocationLongitude !== undefined,
+            defer(() =>
+              Browser.open({
+                url: `https://maps.google.com/maps?q=${proof.geolocationLatitude}, ${proof.geolocationLongitude}`,
+              })
+            )
+          )
+        ),
+        untilDestroyed(this)
+      )
+      .subscribe();
+  }
+
   private share() {
     this.proof$
       .pipe(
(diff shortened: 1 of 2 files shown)


src/app/features/home/capture-tab/capture-details/capture-details.page.ts
```diff
@@ -166,6 +166,26 @@ export class CaptureDetailsPage {
       .subscribe();
   }
 
+  openMap() {
+    return this.proof$
+      .pipe(
+        concatMap(proof =>
+          iif(
+            () =>
+              proof.geolocationLatitude !== undefined &&
```
shc261392, line 175: For some old data (confirmed to exist on Production account), the `proof.geolocationLatitude` and `proof.geolocationLongitude` is string type `'undefined'` and cannot be excluded with this conditional statement.

It also displays incorrectly in Capture detail page (`undefined, undefined` instead of `Location Not Disclosed`)

This kind of case should be handled

src/app/features/home/capture-tab/capture-details/capture-details.page.ts
```diff
@@ -166,6 +166,26 @@ export class CaptureDetailsPage {
...
+              proof.geolocationLongitude !== undefined,
+            defer(() =>
+              Browser.open({
+                url: `https://maps.google.com/maps?q=${proof.geolocationLatitude}, ${proof.geolocationLongitude}`,
```
shc261392, line 179:
```suggestion
                url: `https://maps.google.com/maps?q=${proof.geolocationLatitude},${proof.geolocationLongitude}`,
```

Drop the space between latitude and longitude, it's not necessary and it prevents the iOS App from opening the url successfully. (It seems the iOS browser won't do the url encoding automatically. Instead, it reports "Invalid URL" and refuses to open the link.)

shc261392 (changes requested):

shc261392 (approved):

Issue #388, opened 2020-12-21
Can reset password


scott-dt:
## Concept
Can reset password

PR #549, opened 2021-02-20
Implement the password-resetting dialog.

seanwu1105:
See #388.

### DIFF
--- a/src/app/features/login/login.page.html
+++ b/src/app/features/login/login.page.html
@@ -15,6 +15,15 @@
     </button>
   </form>
 
+  <button
+    (click)="forgetPassword()"
+    class="expand"
+    color="primary"
+    mat-stroked-button
+  >
+    {{ t('forgetPassword') }}
+  </button>
+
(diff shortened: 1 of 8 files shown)


shc261392 (changes requested):
The code looks good. But somehow I feel a bit unsure about resetting the password inside profile page. I'm not followed up with the feature discussion so I may be wrong, but it seems like the password-resetting feature is for the users who forget their password?

If that is the case, the button should be placed at the login page.

seanwu1105:
@shc261392 According to the discussion on Slack:

> @seanwu1105 
> @tammyyang  FR Can Reset Password 是要做功能在 profile page 還是 login page？或者是兩者都要？
> 
> @tammyyang 
> @seanwu1105  這個 sprint 做 login 就好

I have moved the implmentation to the login page and add a input dialog for users to enter their registered email. This FR has been tested on Exodus 1 and Brave browser.

shc261392 (approved):

PR #551, opened 2021-02-23
Improve register performance

seanwu1105:
Improve the register performance by

- Using offline preview image on sending-post-capture page.
- Checking if `sharable_copy` is available from the backend before showing share moment option on capture-details page.

The PR has been tested on Brave browser and Exodus 1.


shc261392 (approved):
Querying proof & loading the image every time the page redirects (For example, from `capture-details` to `sending-post-capture`, even when the resources are loaded on both page, they could not share it) seems not so good to me, but I understand that there is no simple way to fix it now. I'll open a feature request for it.

PR #569, opened 2021-03-11
Remove geolocation icon on PostCaptureTabComponent.

seanwu1105:
According to the feedback from @tammyyang on Sprint and mid-sprint demo, we do not need to show the geolocation icon on PostCapture tab.


shc261392 (approved):
Tested on iPhone 12 Pro Max simulator

PR #576, opened 2021-03-16
Get Thumbnail src with URI instead of DataURL

seanwu1105:
Use `Filesystem.getUri()` method in Capacitor plugin to handle thumbnail request. The method does not work in browser, and thus need to check the environment before request URI. 

By doing so, we do not need to serialize/deserialize the image file to base64, which reduces some memory usage.

This PR has been tested on Brave browser and Exodus 1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "material-design-icons-iconfont": "^6.1.0",
     "merge-images": "^2.0.0",
     "rxjs": "~6.6.3",
+    "safe-pipe": "^1.0.3",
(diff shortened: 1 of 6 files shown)


bafu (approved):
The PR looks good to me. :+1: 

One suggestion: adding a comment for `getThumbnailUrl` to describe the benefit of using `Filesystem.getUri()`

shc261392 (approved):
Code looks good, tested on iOS.

PR #581, opened 2021-03-18
Refresh PostCaptures when accept transaction

seanwu1105:
See #582.

- Fetch all PostCaptures on showing home page.
- Make PostCaptures stream reactive, emitting multiple values instead of one time event.
- Rename `fetchPostCaptures$` to `getPostCaptures$`, which indicating returning multiple values.


src/app/shared/services/dia-backend/transaction/dia-backend-transaction-repository.service.ts
```diff
@@ -85,6 +87,11 @@ export class DiaBackendTransactionRepository {
           {},
           { headers }
         )
+      ),
+      tap(() =>
+        this.assetRepositroy.refreshPostCaptures({
+          reason: `${DiaBackendTransactionRepository.name}.${this.accept$.name}`,
```
shc261392, line 93: What is the reason to pass the class/function name as `{ reason: string }` object to AssetRepositoryService? It looks like it only serves for debugging purpose

shc261392 (changes requested):
The renaming parts are quite good, though there are parts that I have question for

src/app/shared/services/dia-backend/transaction/dia-backend-transaction-repository.service.ts
seanwu1105: @shc261392 Yes, the `{ reason: string }` object is only for debugging. I think it could help the developer to trace whether and which service trigger the refreshing. I can add comments for this argument. If you think this argument is unnecessary, I can remove it of course.

shc261392: If it would make developing and debugging easier during the development phase that the PostCapture-refreshing feature constantly needs to be updated or modified, it's ok to have it. But the call signature is not very clear for the caller to understand what it should put in the function arguments.

If we would like to the caller to pass its class/function name, taking arguments like `{ callerClassName: string, callerMethodName: string }` would make more sense.

seanwu1105: @shc261392 I only added the comments for the `reason` argument as `refreshPostCaptures` method might not always be called by another singleton object or instance. Regarding flexibility and readability, I think using a plain text with comment only should be enought for debugging.

shc261392 (approved):

PR #588, opened 2021-03-22
Open store with in-app browser

seanwu1105:
- Implement store button on tab navbar.
- Open store with in-app browser.
- Apply better validation on geolocation icons.


shc261392 (approved):
Tested on iPhone 12 Pro Max simulator and iPhone 7 Plus

Simulator

Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-23 at 12 44 07

iPhone 7 Plus

IMG_0027

PR #596, opened 2021-03-24
Avoid fetching expired PostCaptures created before onboarding timestamp

seanwu1105:
See #595.

### DIFF
--- a/src/app/features/home/home.page.ts
+++ b/src/app/features/home/home.page.ts
@@ -61,7 +61,7 @@ export class HomePage {
   }
 
   private async onboardingRedirect() {
-    if (await this.onboardingService.isOnboarding()) {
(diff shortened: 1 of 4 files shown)


shc261392 (changes requested):
Unexpected behavior (even if choose to skip pre-fetching at the first time opening the app, it will show the upgrading dialog when the second time opening the app)

Can be reproduced on web

Steps to reproduce:
1. Open app with clean data and login
2. Choose `Skip` (pref key `TO_0_15_0` is not set at this moment)
3. Refresh the page, a upgrading dialog will be shown (pref key `TO_0_15_0` will be set after this, and the dialog won't appear again)

shc261392 (approved):

PR #600, opened 2021-03-29
Implement memory cache for PostCaptures.

seanwu1105:
See #554.

According to the discussion on mid-sprint kickoff, implement in-memory cache for PostCaptures to provide better UX.

- Cache PostCapture list response in memory.
- Cache PostCapture `asset_file` in memory.
    - Possible thrashing with high `asset_file` quantity.

This is NOT an ideal cache implementation as we should build a safer and more structural LRU cache by leveraging declarative caching tool kits, such as `ngsw`. However, we do not have enought time for now.

### DIFF
--- a/src/app/shared/services/dia-backend/asset/dia-backend-asset-repository.service.ts
+++ b/src/app/shared/services/dia-backend/asset/dia-backend-asset-repository.service.ts
@@ -19,6 +19,7 @@ import {
 import {
   catchError,
   concatMap,
+  distinctUntilChanged,
   first,
   map,
   pluck,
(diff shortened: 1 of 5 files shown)


src/app/shared/services/dia-backend/asset/dia-backend-asset-repository.service.ts
```diff
@@ -73,15 +95,9 @@ export class DiaBackendAssetRepository {
...
-    return iif(
-      () => options?.limit !== undefined,
-      this.list$({
-        isOriginalOwner: false,
-        orderBy: 'source_transaction',
-        limit: options?.limit,
-        offset: options?.offset,
-      }),
+  getPostCaptures$() {
+    return merge(
```
shc261392, line 99: The order of emissions from input observables are not enforced, therefore it's theoretically possible that the cache emits after the API response emits and that wouldn't be an expected behavior. Since it's very unlikely to happen in real life and even if it happens it won't be a critical issue, I won't request change for it.

shc261392 (approved):

src/app/shared/services/dia-backend/asset/dia-backend-asset-repository.service.ts
seanwu1105: I don't think this will happen. Since the cache will be updated by [`tap(response => this.postCapturesCache$.next(response))*`](https://github.com/numbersprotocol/capture-lite/pull/600/files/b5d20f2bafb1a83df82e1c7955b5f218a231ccd6#diff-70f3e585f461105d08f253b8ed0b479278d855b905c21ea263f5c24418bc618eR110). Thus, the cache will always emit the latest value after the data is fetched from the backend.

shc261392 (approved):

PR #602, opened 2021-03-30
Implement fullscreen image viewer

seanwu1105:
See #597.


shc261392 (approved):
Tested on iOS. Confirmed that the gesture zooming works on iOS physical device.

PR #603, opened 2021-03-30
Update Angular from 10.0 to 11.0

seanwu1105:
The update brings the following critical benefit for the app.

- Improve app start-up performance by latest Ivy engine.
- New features (such as resizable item) in virtual scroll in `@angular/material` and `@angular/cdk` will only be added on version 11 or later.
- [Ionic has fully support Angular 11](https://ionicframework.com/blog/massive-updates-with-version-11-of-angular/).
- Faster testing speed with karma 6 for CI/CD.

Angular 11 also brings a better developer experience.

- Drop deprecated TSLint. No more false positive TSLint warnings and errors showing on VSCode hint. (Need to add ESLint in the future.)
- Provide better compatibility with Angular Language Server on VSCode to avoid false positive error on editing code.
- Provide strong-typing for templates.

Angular 11 does __NOT__ introduce any incompatible behavior to our current app as it is fully compatible with the current major version of Chromium engine. Tested on Brave browser and Exodus 1.


shc261392 (approved):
Tested on iOS

PR #604, opened 2021-03-31
Improve loading capture into image-viewer performance

seanwu1105:
Improve the performance by using native `Filesystem.getUri` instead of data URL when loading original images.

The previous experiments, such as #576, indicates that reading multiple small images with base64 is faster than with `getUri`. However, reading single large image with base64 is much slower than `getUri`.


shc261392 (approved):
Tested on iOS

PR #605, opened 2021-03-31
Make contacts header style consistent with other headers

seanwu1105:
- Force Ionic style mode to be `md`.
- Make the text in `ion-title` to be centered.

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -18,7 +18,7 @@ import { SharedModule } from './shared/shared.module';
     SharedModule,
     BrowserModule,
     BrowserAnimationsModule,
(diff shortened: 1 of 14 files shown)


shc261392 (changes requested):
Simulator Screen Shot - iPhone 12 Pro Max - 2021-04-01 at 10 47 32
Simulator Screen Shot - iPhone 12 Pro Max - 2021-04-01 at 10 47 39

The styling is still not consistent with other headers on newer iPhone.
The header uses `ion-toolbar` component while all the other headers uses `mat-toolbar` component.

shc261392 (approved):

PR #607, opened 2021-04-01
Make network status bar position consistent

seanwu1105:
According to the discussion of demo on 03/31/2021, we need to move the network status bar position below `ion-segment` on `PostCaptureTabComponent`.


shc261392 (approved):
Tested on iOS

PR #616, opened 2021-04-06
Improve performance of loading image thumbnail after capturing.

seanwu1105:
- Implement individual CaptureItemComponent. -4 seconds
- Avoid unnecessary thumbnail reading by checking if proof is
  updated. -10 seconds
- Replace image-blob-reduce with comporessorjs. -1 seconds
- Display original image first when creating thumbnail. -5 seconds

Tested on Brave browser and Exodus 1 with ~250 pictures.


shc261392 (approved):
On iPhone 7 Plus, it takes 3~4 sec to see the grey image skeleton when having about 60 images and while performing upload.

Issue #611, opened 2021-04-01
Integrate series view


scott-dt:
Integrate series view

PR #618, opened 2021-04-07
Connect with series APIs.

seanwu1105:
See #611.

- Get total series count to fetch ALL series with API.
- Read series details only by its ID.
- Display gray blocks if the assets in the collections are not collected. [See discussion](https://dt42-numbers.slack.com/archives/C0161QL0Z4N/p1617771376019100?thread_ts=1617766656.015900&cid=C0161QL0Z4N).
- Display gray blocks if the series does not have cover image. [See discussion](https://dt42-numbers.slack.com/archives/C0161QL0Z4N/p1617777848021500).


shc261392 (approved):
Tested on iOS

PR #623, opened 2021-04-09
Change selectedIndex of home tabs to capture tab after capturing.

seanwu1105:
See [the suggestion on Slack](https://dt42-numbers.slack.com/archives/CS29L7Z5H/p1617950529211300).


shc261392 (approved):
It takes 2-3 seconds to navigate to the Capture tab after the camera is closed on iPhone 7 Plus. The implementation is ok though I'm not sure if the UX could meet the expectation.

Issue #612, opened 2021-04-01
Support video-type asset (60s) - can view and play


scott-dt:
Support video-type asset (60s) - can view and play

PR #627, opened 2021-04-13
Support video view.

seanwu1105:
See #612.

Currently, the app only support video view but creation. The implementation is a little bit dirty (lots of if-else statement :cry: ) as I try to avoid breaking changes on data source.


shc261392 (approved):
Tested on iOS

PR #631, opened 2021-04-19
Show error toasts on HTTP error response

seanwu1105:
See #240.

- Catch most of `HttpErrorResponse`s from DIA backend.
- Add `NUMBERS_STORAGE_TRUSTED_CLIENT_KEY` to the header of `createUser` endpoint.

After we catch most of the "expected" errors, we can introduce crash report without noise exceptions.


src/app/shared/modules/error/error.service.ts
```diff
@@ -0,0 +1,72 @@
...
+              this.toastController
+                .create({ message, duration: 2000, color: 'danger' })
+                .then(toast => toast.present())
+                .then(() => {
+                  console.error(error);
+                  resolve(error);
+                })
+            )
+        ),
+        concatMapTo(NEVER)
```
shc261392, line 62: `NEVER` never completes the observable, but `BlockingActionService.run$` requires the observable to complete in order to dismiss loading. There will be infinite loading if an error occurs during a blocking action.

shc261392 (changes requested):

PR #680, opened 2021-05-10
Support video recording.

seanwu1105:
See #676.

- Unlimited recording time due to implementation limitation. See https://github.com/w3c/html-media-capture/issues/23 for details.
- Only support `video/mp4` and `video/quicktime` type with H.264 encoding.
- HTML media capture support list: https://caniuse.com/html-media-capture
- Update Android Studio environment configurations.
- Tested on Exodus 1 with latest webview version.


shc261392 (approved):
Tested on iOS

PR #683, opened 2021-05-12
Support large media asset.

seanwu1105:
- Use `capacitor-blob-writer` to write large file into native filesystem.
- Use `HttpClient` to read large file into Angular.
- Avoid using `Capacitor.Filesystem.readFile` and `Capacitor.Filesystem.writeFile` if possible.
- Create video thumbnail with video URL instead of base64 string.

Tested on Exodus 1 with 20 seconds video recording.


shc261392 (approved):
Tested on iOS

bafu:
The PR seems to have a side effect: user will fail to restore the videos

bafu  6:52 PM
@Sean Wu video feature 好像會影響初次登入的 restore，我會一直看到 Fail to pull old Captures dialog

Sean Wu  6:52 PM
@bafu 我也有遇到，這個問題還要研究一下，因為它的確可以用 curl 取得。

I will wait for Sean's investigation of this issue.

bafu:
This PR is merged w/o code updating because the [root cause is related to the backend](https://dt42-numbers.slack.com/archives/C3CHPJTS4/p1620895310147000?thread_ts=1620791473.053000&cid=C3CHPJTS4).

PR #692, opened 2021-05-24
Use ETH wallet signature

seanwu1105:
Replace ECDSA SECP256R1 signature with ETH wallet sign.

Tested on Brave and Exodus 1.


shc261392 (approved):
Tested on iOS

PR #752, opened 2021-06-23
fix color display in any theme

CS6:
image
image
image
image


shc261392 (approved):
The change looks fine. I'm accepting this PR but I did not test them on the described environments.

PR #775, opened 2021-07-02
Feature fix long video crash

shc261392:
Use sha256sum instead of file content base64 as object key in signed message.


bafu:

bafu (approved):

PR #777, opened 2021-07-06
Feature improve home loading time

shc261392:
Improve the initial loading time and priority of Home page
- Use file URL instead of data URI for thumbnails to truly enable lazy loading
- Preload launcher.svg
- Remove unneeded remote Material Icon source
- Replace ion-icon which loads slowly at App startup with mat-icon


bafu (approved):

bafu:

PR #784, opened 2021-07-15
Send user's auth token along with URL when opening Capture Club from App

vincent10400094:
Please check that whether I misused the RxJS operators, it's my first time using them.

### DIFF
--- a/src/app/features/home/details/details.page.ts
+++ b/src/app/features/home/details/details.page.ts
@@ -6,7 +6,14 @@ import { ActionSheetController } from '@ionic/angular';
 import { ActionSheetButton } from '@ionic/core';
 import { TranslocoService } from '@ngneat/transloco';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
-import { combineLatest, defer, EMPTY, Observable, ReplaySubject } from 'rxjs';
+import {
+  combineLatest,
+  defer,
+  EMPTY,
+  Observable,
+  of,
+  ReplaySubject,
+} from 'rxjs';
 import {
   catchError,
   concatMap,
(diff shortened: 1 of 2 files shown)


src/app/features/home/home.page.ts
```diff
@@ -197,9 +198,17 @@ export class HomePage {
...
-    });
+    return this.diaBackendAuthService.token$
+      .pipe(
+        first(),
+        switchMap(token => {
+          return Browser.open({
+            url: `https://captureclub.cc/?token=${token}`,
+            toolbarColor: '#564dfc',
+          });
+        }),
```
shc261392, line 209:
```suggestion
        switchMap(token =>
          Browser.open({
            url: `https://captureclub.cc/?token=${token}`,
            toolbarColor: '#564dfc',
          })
        ),
```

Could use implicit return for single line arrow function.

src/app/features/home/home.page.ts
```diff
@@ -197,9 +198,17 @@ export class HomePage {
 
   // eslint-disable-next-line class-methods-use-this
   async openCaptureClub() {
```
shc261392, line 200:
```suggestion
  openCaptureClub() {
```
Keeping `async` keyword might be confusing since the function is not supposed to return Promise anymore.

src/app/features/home/home.page.ts
```diff
@@ -197,9 +198,17 @@ export class HomePage {
...
-    return Browser.open({
-      url: 'https://captureclub.cc/',
-      toolbarColor: '#564dfc',
-    });
+    return this.diaBackendAuthService.token$
```
shc261392, line 201:
```suggestion
    this.diaBackendAuthService.token$
```
The function now returns the subscription, but we'd almost never need to get the return value of a subscription. Better return void instead.

src/app/features/home/details/details.page.ts
```diff
@@ -379,16 +386,21 @@ export class DetailsPage {
   }
 
   private openCaptureClub() {
-    return this.activeDetailedCapture$
+    return combineLatest([
+      this.activeDetailedCapture$,
+      this.diaBackendAuthService.token$,
+    ])
```
shc261392, line 392:
```suggestion
    return combineLatest([
      this.activeDetailedCapture$.pipe(
        concatMap(
          activeDetailedCapture => activeDetailedCapture.diaBackendAsset$
        ),
        isNonNullable()
      ),
      this.diaBackendAuthService.token$,
    ])
```

It appears that the first `concatMap` after `first()` operator could be removed. The whole pipe would be a bit simpler this way.

shc261392 (changes requested):
There are a few places that could be changes. They're minor but should improve the coding style a bit. It's open for discussion if you have any other ideas.

I didn't test out this changes, so if you follow these changes please test it again to make sure the functionality doesn't change.

shc261392 (approved):

PR #785, opened 2021-07-19
Make file input element for video hidden and removed when user cancle…

vincent10400094:
1. When user cancel to upload video, the newly created file input element will be removed now
2. File input element is added with attribute "hidden", so that users will not see the unintended input element


shc261392 (approved):
Tested on Android and iOS device.

PR #787, opened 2021-07-22
Implement phone verification and email verification features

vincent10400094:
- Fix minor translation issue in figma
- For email verification, it currently sends account activation link instead of verification code
- If phone verification code is verified successfully, the verification page will be closed automatically after 3 seconds

### DIFF
--- a/src/app/features/profile/email-verification/email-verification.page.ts
+++ b/src/app/features/profile/email-verification/email-verification.page.ts
@@ -1,8 +1,17 @@
 import { Component } from '@angular/core';
 import { TranslocoService } from '@ngneat/transloco';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
-import { defer, timer } from 'rxjs';
-import { catchError, first, map, switchMap, take, tap } from 'rxjs/operators';
+import { timer } from 'rxjs';
+import {
+  catchError,
+  concatMap,
+  finalize,
+  first,
+  map,
+  switchMap,
+  take,
+  tap,
+} from 'rxjs/operators';
 import { BlockingActionService } from '../../../shared/blocking-action/blocking-action.service';
 import { DiaBackendAuthService } from '../../../shared/dia-backend/auth/dia-backend-auth.service';
 import { ErrorService } from '../../../shared/error/error.service';
@@ -16,7 +25,6 @@ import { ErrorService } from '../../../shared/error/error.service';
 export class EmailVerificationPage {
   hasSentEmailVerification = false;
   secondsRemained = 0;
-  readonly RESEND_INTERVAL: number = 60;
 
   readonly emailVerified$ = this.diaBackendAuthService.emailVerified$;
   readonly email$ = this.diaBackendAuthService.email$;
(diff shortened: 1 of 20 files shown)


src/app/features/profile/email-verification/email-verification.page.ts
```diff
@@ -0,0 +1,58 @@
...
+      ),
+      catchError((err: unknown) => this.errorService.toastError$(err)),
+      tap(() => {
+        this.hasSentEmailVerification = true;
+        timer(0, ONE_SECOND)
+          .pipe(
+            take(this.RESEND_INTERVAL),
+            map(i => this.RESEND_INTERVAL - i - 1)
+          )
+          .subscribe(val => {
```
shc261392, line 46: Nested subscription in RxJS is an [anti-pattern](https://angular-checklist.io/default/checklist/rxjs/Z1eFwa9) best avoiding. Almost under no circumstances would nested subscription be an ideal implementation when writing RxJS. In other words, if you find yourself doing nested subscription, there is almost always a better way to do it.

Here's a few hint on how to rewrite it to avoid nested subscription:

```
const RESEND_COOLDOWN_TICKS = 60;
const TICK_INTERVAL = 1000;
const countdown$ = timer(0, TICK_INTERVAL).pipe(
  take(RESEND_COOLDOWN_TICKS),
  map(tick => RESEND_COOLDOWN_TICKS - tick - 1),
  tap(cooldown => this.secondsRemained = cooldown),
  finalize(() => this.secondsRemained = 0)
);
```

```
return this.blockingActionService
  .run$(action$, {
      message: this.translocoService.translate('message.pleaseWait'),
  })
    .pipe(
      switchMap(() => countdown$),
      untilDestroyed(this)
    )
    .subscribe();
```

In my opinion `switchMap` is safer here since it guarantees the countdown to be reset every time the action is run even if the button is not disabled.

If the button is disabled during countdown, it is effectively the same as `concatMap` and we could use `concatMap` instead.

src/app/features/profile/email-verification/email-verification.page.ts
```diff
@@ -0,0 +1,58 @@
...
+    private readonly diaBackendAuthService: DiaBackendAuthService,
+    private readonly errorService: ErrorService,
+    private readonly translocoService: TranslocoService
+  ) {}
+
+  async sendEmailVerification() {
+    const ONE_SECOND = 1000;
+    const action$ = defer(() => this.email$).pipe(
+      first(),
+      switchMap(email =>
```
shc261392, line 35: When there's no need to cancel previous inner observable (`resendActivationEmail$`) when source observable (`email$`) emits new value, use `concatMap` instead.

It's nearly effectively the same in this case, but when not intentionally want to use the cancel capability of `switchMap`, `concatMap` is a more natural choice.

src/app/features/profile/phone-verification/phone-verification.page.ts
```diff
@@ -0,0 +1,186 @@
...
+      .pipe(
+        catchError((err: unknown) => this.errorService.toastError$(err)),
+        tap(() => {
+          this.hasSentPhoneVerification = true;
+          timer(0, ONE_SECOND)
+            .pipe(
+              take(this.RESEND_INTERVAL),
+              map(i => this.RESEND_INTERVAL - i - 1)
+            )
+            .subscribe(val => {
```
shc261392, line 125: See the comment in email verification part.

src/app/features/profile/phone-verification/phone-verification.page.ts
```diff
@@ -0,0 +1,186 @@
...
+      )
+      .pipe(
+        catchError((err: unknown) => this.handleVerificationError$(err)),
+        tap(() => {
+          this.snackBar.open(
+            this.translocoService.translate('message.verificationSuccess')
+          );
+          timer(THREE_SECONDS)
+            .pipe(first())
+            .subscribe(() => {
```
shc261392, line 153: See the comment in email verification part.

src/app/features/profile/phone-verification/phone-verification.page.ts
```diff
@@ -0,0 +1,186 @@
...
+      );
+    return this.blockingActionService
+      .run$(action$, {
+        message: this.translocoService.translate('message.pleaseWait'),
+      })
+      .pipe(untilDestroyed(this))
+      .subscribe();
+  }
+
+  private handleVerificationError$(err: unknown) {
```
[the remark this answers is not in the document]
shc261392: There is also a `throttled` error that could possibly happen when the user is sending verification phone number or code over 5 times per minute. This is achievable by leaving the page and go back to the page again (the current button lock status and cooldown would not persist after component destruction).

The priority is low, however, since it only happens when user tries to intentionally abuse the mechanism of App, and the effect is not permanent. (after 1 minute the API will stop responding will `throttled` error) Most user would probably never meet this API error.

shc261392 (changes requested):
There are some changes that need to be addressed. Please see the code comments.

shc261392 (approved):
Everything looks good 🎉

Issue #779, opened 2021-07-12
Create a stable and reviewed version of signed message


shc261392:
## Concept

Currently, there are two version of signed messages created from Capture App.

Pre 0.34.2 example:

```
{
  "indexedAssets":{
    "<the-file-content-encoded-as-base64>":{
      "mimeType":"video/mp4"
    }
  },
  "truth":{
    "providers":{
      "CapacitorFactsProvider":{
        "APP_BUILD":"152",
        "APP_ID":"io.numbersprotocol.capturelite",
        "APP_NAME":"Capture",
        "APP_VERSION":"0.32.2",
        "BATTERY_LEVEL":0.9599999785423279,
        "DEVICE_NAME":"EXODUS 1",
        "DISK_FREE":412057600,
        "DISK_TOTAL":4225581056,
        "GEOLOCATION_LATITUDE":25.005851,
        "GEOLOCATION_LONGITUDE":121.4586974,
        "IS_CHARGING":true,
        "IS_VIRTUAL":false,
        "MANUFACTURER":"HTC",
        "MEM_USED":5414752,
        "OPERATING_SYSTEM":"android",
        "OS_VERSION":"9",
        "PLATFORM":"android",
        "UUID":"16eb24e0fa1f8a28"
      }
    },
    "timestamp":1625213302097
  }
}
```

0.34.2 or later example:

```
{
  "indexedAssets":{
    "47af037408a25364f8d18b133fa96ba4afb2b60e53d5329ee629e2a28a8e2bcd":{
      "mimeType":"video/mp4"
    }
  },
  "truth":{
    "providers":{
      "CapacitorFactsProvider":{
        "APP_BUILD":"152",
        "APP_ID":"io.numbersprotocol.capturelite",
        "APP_NAME":"Capture",
        "APP_VERSION":"0.32.2",
        "BATTERY_LEVEL":0.9599999785423279,
        "DEVICE_NAME":"EXODUS 1",
        "DISK_FREE":412057600,
        "DISK_TOTAL":4225581056,
        "GEOLOCATION_LATITUDE":25.005851,
        "GEOLOCATION_LONGITUDE":121.4586974,
        "IS_CHARGING":true,
        "IS_VIRTUAL":false,
        "MANUFACTURER":"HTC",
        "MEM_USED":5414752,
        "OPERATING_SYSTEM":"android",
        "OS_VERSION":"9",
        "PLATFORM":"android",
        "UUID":"16eb24e0fa1f8a28"
[...]

PR #795, opened 2021-07-23
Feature signed message

shc261392:
- Generate and save signature as discussed in #779 
- Trigger updating new signature and signed metadata to DIA backend on these conditions:
  - Upgrading from 0.35.0 or lower version
  - After pre-fetch is completed


bafu (approved):

bafu:

PR #797, opened 2021-07-23
Feature dark mode android workaround

vincent10400094:
- Apply [this workaround](https://github.com/ionic-team/capacitor/discussions/1978#discussioncomment-708439) to force activating dark mode on Android.
- Fix buttons display issue in login page and signup page

### DIFF
--- a/android/app/src/main/java/io/numbersprotocol/capturelite/MainActivity.java
+++ b/android/app/src/main/java/io/numbersprotocol/capturelite/MainActivity.java
@@ -8,8 +8,39 @@ import com.getcapacitor.Plugin;
 import com.equimaps.capacitorblobwriter.BlobWriter;
 
 import java.util.ArrayList;
+import android.content.res.Configuration;
+import android.webkit.WebSettings;
 
 public class MainActivity extends BridgeActivity {
+  void setDarkMode() {
+    // WORKAROUND: Android device doesn't support media query for prefers-color-scheme
+    // @reference: https://github.com/ionic-team/capacitor/discussions/1978#discussioncomment-708439
+    int nightModeFlags = getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK;
+    WebSettings webSettings = this.bridge.getWebView().getSettings();
+    if (nightModeFlags == Configuration.UI_MODE_NIGHT_YES) {
+      if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
+        // As of Android 10, you can simply force the dark mode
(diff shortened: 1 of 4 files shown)


vincent10400094:
## Result

combine_images-3

## Know issue

 Some icons can not be seen in Android dark mode (but can be seen in iOS dark mode)

1. Capture details page icons
2. Post capture tab icons
combine_images-5

android/app/src/main/java/io/numbersprotocol/capturelite/MainActivity.java
```diff
@@ -8,8 +8,39 @@
...
+      if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
+        // As of Android 10, you can simply force the dark mode
+        webSettings.setForceDark(WebSettings.FORCE_DARK_ON);
+      }
+    } else {
+      if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
+        webSettings.setForceDark(WebSettings.FORCE_DARK_OFF);
+      }
+    }
+  }
```
shc261392, line 30:
```suggestion
    final int nightModeFlags = getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK;
    WebSettings webSettings = this.bridge.getWebView().getSettings();
    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
        final int forceDark;
        forceDark = nightModeFlags == Configuration.UI_MODE_NIGHT_YES
                ? WebSettings.FORCE_DARK_ON
                : WebSettings.FORCE_DARK_OFF;
        webSettings.setForceDark(forceDark);
    }
```

The readability might be better this way. (Repeating the conditional statement looks a bit confusing)

android/app/src/main/java/io/numbersprotocol/capturelite/MainActivity.java
```diff
@@ -8,8 +8,39 @@
...
+
+  @Override
+  public void onStart() {
+    super.onStart();
+    setDarkMode();
+  }
```
shc261392, line 36:
```suggestion
```

`onResume()` is always called before an Activity is running in the foreground, including first time opening the app or resuming from a background paused state, see the flowchart part of [Activity Lifecycle](https://developer.android.com/guide/components/activities/activity-lifecycle).

In that case, setting dark mode in one lifecycle hook should be enough. `onResume()` would be better since it is always triggered.

shc261392 (changes requested):

shc261392 (approved):

PR #840, opened 2021-08-20
Feature fix darkmode display issues

vincent10400094:
## Overview

Dark mode of Capture was first implemented with media query (`prefer-color-scheme: dark`). However, it's not compatible with Android, even with [this workaround](https://github.com/ionic-team/capacitor/discussions/1978#discussioncomment-24217), the media query is not activated properly (Android forces dark theme by inverting the colors instead of using the dark theme we defined).

Alternatively, the [CSS class fallback](https://ionicframework.com/docs/theming/dark-mode#css-class-fallback) seems to work on both iOS and Android. And here are the changes I made:

1. Add the `dark` class to `document.body` when Capture starts and system dark mode is detected
2. Add dark theme for material theme globally. By doing so, most of material components can be turned into dark mode automatically, and there's no need to specify dark mode appearance for each component

Ideally, dark mode can work on both platform, and issues #835 #836 should be fixed now.

## Known issue
- Due to technical issues, Capture only detects dark mode when starting. So if you change the system preference, the App has to be restarted to take effect.


shc261392:
## Current version on Play Store
photo_2021-08-23 10 31 02

## The PR version
photo_2021-08-23 10 31 00

The status bar is wider than the Play Store version. Is it intended or a side effect?

Phone: Nokia 6.1
OS: Android 10

vincent10400094:
It's not intended. I think it's caused by "material dark theme". Let me see whether it can be fixed.

vincent10400094:
The height of toolbar should be consistent now.

shc261392 (commented):
Looks good to me, tested on Nokia 6.1 (Android 10).

shc261392 (approved):

PR #845, opened 2021-08-30
Share cai-injected photo instead of sharable copy

vincent10400094:
#844 
- Share CAI-injected photo so that photos can speak for themselves
- Shared photo can be parsed by [media-reader](https://authmedia.net/media-reader) only when sharing them as "files" (e.g. sharing with gmail/gdrive, etc) 
    - iOS: Share as "saving to files" instead of "saving to photos"
    - APPs like instagram, facebook will compress the photo and the photo can not be parsed with our [media-reader](https://authmedia.net/media-reader)


shc261392 (approved):
Looks good to me. Tested on iPhone 7 Plus.

PR #879, opened 2021-09-06
Fix duplicate paste issue (verification code form for phone verification)

vincent10400094:
#839 

#### Overview

- Fix duplicate paste issue (verification code form for phone verification)
- Fix homepage sidebar display issue (before & after):

combine_images

#### Note:
- For `formly-form`, there's duplicate paste issue on iOS if using input type `tel`


shc261392 (approved):
Code looks good to me. Side menu fix tested on iPhone.
Haven't test the iOS code paste issue fix since I can't put my SIM card to my iPhone.

Issue #882, opened 2021-09-08
Can see NUM (CAPT) balance in Capture App


scott-dt:
Can see NUM (CAPT) balance in Capture App **fallow UI #881**

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1201059864610283) by [Unito](https://www.unito.io)

tammyyang:
Done, close issue.

PR #886, opened 2021-09-13
feat: display NUM token balance in profile page

vincent10400094:
#882 
Overview
- Display NUM(CAPT) token balance in profile page
- Balance will be refreshed when opening the profile page
- When the APP get offline, the balance will be displayed as text `--offline--`

Note
- Token is called `NUM` in UI/UX design but called `CAPT` currently (deployed on Ropsten network)


shc261392 (approved):
Everything looks great

PR #916, opened 2021-09-27
fix: use text type to prevent incorrect code length validation

vincent10400094:
Note

The behavior of all types being used as verification code:

- `tel`: keyboard: only digits, issue: double-paste
- `number`: keyboard: digits + punctuation marks, issue: incorrect code validation
- `text`: keyboard: normal alphabet keyboard (need to switch), issue: N/A


shc261392 (approved):
If the alphabet keyboard or allowing user to input alphabet becomes an UX issue, maybe we could bind keypress event to deal with it. But I think it's fine for now.

Issue #898, opened 2021-09-24
Can see NFT address and token ID


scott-dt:
Can see NFT address and token ID

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1201059864319011) by [Unito](https://www.unito.io)

tammyyang:
Migrate to asana <> unito sync, close the issue.

PR #975, opened 2021-10-20
Feature display NFT token and caption

vincent10400094:
#909 #898 

- Can see Capture caption (if there is)
- Can see NFT token information (if there is)
    - Display token type (`TT721` for thundercore token, `ERC721` for ethereum token)
    - Display token ID
    - With explorer link (viewblock.io for thundercore, etherscan.io for ethereum)


shc261392 (approved):
Code looks good to me and I'll accept it directly.

PR #1050, opened 2021-11-23
Feature fix v3 migration issue

vincent10400094:
- Skip migration if there is no previous version (i.e. when the APP is installed and logged in at the first time)
- Fix Capture logo display in signup page


shc261392 (approved):
Looks good to me

PR #1098, opened 2021-12-08
fix(AndroidManifest.xml): add IMAGE_CAPTURE query intent

vincent10400094:
- Fix `Unable to resolve camera activity` issue on Pixel 5 (Android 11, API level 30)

Reference: [bug: Android 11 - Unable to resolve camera activity #3627](https://github.com/ionic-team/capacitor/issues/3627#issuecomment-741703368)


shc261392 (approved):
The fix seems legit. I don't have any device that could reproduce this issue so I'll accept the PR directly.

PR #1110, opened 2021-12-14
add(actions-dialog): add action parameter selection dialog

vincent10400094:
#1049 

* Use `mat-dialog` instead of `ionic alert` to support multiple types of input
  * text input
  * number input
  * dropdown (select)
* Display more information about each input
* Checking the input when submitting the form

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1201515367303313) by [Unito](https://www.unito.io)
┆Created By: Tammy Yang

### DIFF
--- a/src/app/features/home/details/actions/actions.page.ts
+++ b/src/app/features/home/details/actions/actions.page.ts
@@ -5,7 +5,7 @@ import { ActivatedRoute } from '@angular/router';
 import { TranslocoService } from '@ngneat/transloco';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { combineLatest } from 'rxjs';
-import { catchError, concatMap, map, tap } from 'rxjs/operators';
+import { catchError, concatMap, first, map, tap } from 'rxjs/operators';
 import { ActionsDialogComponent } from '../../../../shared/actions/actions-dialog/actions-dialog.component';
 import {
   Action,
@@ -14,6 +14,7 @@ import {
 import { BlockingActionService } from '../../../../shared/blocking-action/blocking-action.service';
 import { DiaBackendAuthService } from '../../../../shared/dia-backend/auth/dia-backend-auth.service';
 import { ErrorService } from '../../../../shared/error/error.service';
+import { isNonNullable } from '../../../../utils/rx-operators/rx-operators';
 
 @UntilDestroy()
 @Component({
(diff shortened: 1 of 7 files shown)


src/app/features/home/details/actions/actions.page.ts
```diff
@@ -51,39 +51,20 @@ export class ActionsPage {
...
+                  params: params,
+                },
+              });
+              dialogRef.afterClosed().subscribe(data => {
+                if (data !== undefined) {
+                  const body = { ...data, token: token, cid: id };
+                  return this.sendAction(action, body);
+                }
+              });
+              resolve();
```
shc261392, line 67: Reasons to make changes:
1. Nested subscription is anti-pattern
2. Wrapping the anonymous function with Promise here is redundant
3. Subscriptions never complete if not leaving this page

The thing we want to do here can be break down into:
1. Open a dialog and get a `dialogRef` object (synchronous)
2. Subscribe to `dialogRef.afterClose()` observable and perform actions (asynchronous)
3. Call `sendAction()` (synchronous)

After breaking down, we could see that there is only one asynchronous function call here, so this will be what we return in the callback for `concatMap` operator. If we return the observable, the observable will be subscribed by the outmost `subscribe()`, so there's no need to add another subscribe inside. Here we got 1 and 2 covered. And we can filter out `undefined` with the helper operator `isNonNullable`.
To perform a synchronous side effect in RxJS pipe, we could use `tap` operator. Placing the `isNonNullable` and `tap` operator in the pipe of inner observable or to the position after `concatMap` does not matter in this case.

Also, this subscription never completes unless completed by `untilDestroyed` by leaving the page. Add a `first()` operator at the start of the pipe will ensure this subscription completes after emitting once. Where to place `first()` does matter, since the subscription won't emit value when cancelling (because `undefined` will be filtered by `isNonNullable()`), so placing `first()` at the end of the pipe will make the subscription still never complete when action is cancelled.

```
        first(),
        concatMap(([params, token, id]) => {
          const dialogRef = this.dialog.open<ActionsDialogComponent>(
            ActionsDialogComponent,
            {
              disableClose: true,
              data: {
                action: action,
                params: params,
              },
            }
          );
          return dialogRef.afterClosed().pipe(
            isNonNullable(),
            tap(data =>
              this.sendAction(action, { ...data, token: token, cid: id })
            )
          );
        }),
```

shc261392 (changes requested):

shc261392 (approved):
Everything looks good

PR #1171, opened 2022-01-11
feat(profile.page): do not show user's private key directly

vincent10400094:
- User can see their private key after confirming with the warnings

Current UX flow:
1. Click "Export Private Key" button
2. Show the warnings. If proceed, go to step 3. Otherwise, the flow is ended
3. Show the private key, and user can copy to clipboard


shc261392 (approved):
Looks great. Tested on browser, doesn't see any problem.

PR #1188, opened 2022-01-19
ci(.github/workflows/Dev-test.yml): add slack notification when devel…

b07902011:
…op update

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1201686537596486) by [Unito](https://www.unito.io)
┆Created By: Tammy Yang


shc261392 (commented):
We already have `build` workflow which is triggered upon every push and pull request. Maybe by adding a slack notification stage to the last part we wouldn't be needing another separated workflow?

Might be able to use something like `if: github.ref == 'refs/heads/develop'` in the slack job stage

shc261392 (approved):
Looks good

PR #1269, opened 2022-02-14
feat(dia-backend-wallet.service.ts): show custody wallet balance

ruizehung:
Update the api request url endpoint in `getManagedWallet$` to show NUM balance from custody/managed wallet instead of Capture wallet.

## Test Plan
Dev storage backend
image

Prod storage backend
image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1201813627329196) by [Unito](https://www.unito.io)
┆Created By: Tammy Yang

### DIFF
--- a/src/app/shared/dia-backend/wallet/dia-backend-wallet.service.ts
+++ b/src/app/shared/dia-backend/wallet/dia-backend-wallet.service.ts
@@ -56,7 +56,7 @@ export class DiaBackendWalletService {
     );
   }
 
-  getManagedWallet$() {
+  getNumWallet$() {
     return defer(() => this.authService.getAuthHeaders()).pipe(
       concatMap(headers => {
         return this.httpClient.get<DiaBackendWallet>(
@@ -68,7 +68,7 @@ export class DiaBackendWalletService {
   }
 
   syncCaptBalance$() {
-    return this.getManagedWallet$().pipe(
+    return this.getNumWallet$().pipe(
       concatMap(diaBackendWallet =>
         forkJoin([
           this.preferences.setNumber(
--- a/src/app/shared/dia-backend/wallet/dia-backend-wallet.service.ts
+++ b/src/app/shared/dia-backend/wallet/dia-backend-wallet.service.ts
@@ -60,7 +60,7 @@ export class DiaBackendWalletService {
     return defer(() => this.authService.getAuthHeaders()).pipe(
       concatMap(headers => {
         return this.httpClient.get<DiaBackendWallet>(
-          `${BASE_URL}/api/v3/wallets/managed-wallet/`,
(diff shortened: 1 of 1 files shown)


shc261392 (changes requested):
We have an endpoint named `/wallets/managed-wallet/`. To avoid ambiguity, the function name `getManagedWallet$` should be changed to `getNumWallet$` after confirming `getManagedWallet$` is not used in any other places.

ruizehung:
> We have an endpoint named `/wallets/managed-wallet/`. To avoid ambiguity, the function name `getManagedWallet$` should be changed to `getNumWallet$` after confirming `getManagedWallet$` is not used in any other places.

I think it's better for the function to be descriptive about what it does and abstract away the underlying details like which api endpoint it is calling. 

Is the `/wallets/num-wallet/` endpoint a temporary solution? Are we going to keep calling `/wallets/num-wallet/` in the future (and will not call `/wallets/managed-wallet/` anymore)? Or after the backend migration completes, we are going to swtich back to calling `/wallets/managed-wallet/`?

If we will keep calling `/wallets/num-wallet/` in the future, I think it makes sense to rename `getManagedWallet$` to  `getNumWallet$` and intrepret it as a function that gets the wallet used for NUM payment, which can be either asset wallet or managed wallet depending on how the backend is configured. Otherwise, I think it's better to keep its name as `getManagedWallet$`.

What do you think?

shc261392:
> > We have an endpoint named `/wallets/managed-wallet/`. To avoid ambiguity, the function name `getManagedWallet$` should be changed to `getNumWallet$` after confirming `getManagedWallet$` is not used in any other places.
> 
> I think it's better for the function to be descriptive about what it does and abstract away the underlying details like which api endpoint it is calling.
> 
> Is the `/wallets/num-wallet/` endpoint a temporary solution? Are we going to keep calling `/wallets/num-wallet/` in the future (and will not call `/wallets/managed-wallet/` anymore)? Or after the backend migration completes, we are going to swtich back to calling `/wallets/managed-wallet/`?
> 
> If we will keep calling `/wallets/num-wallet/` in the future, I think it makes sense to rename `getManagedWallet$` to `getNumWallet$` and intrepret it as a function that gets the wallet used for NUM payment, which can be either asset wallet or managed wallet depending on how the backend is configured. Otherwise, I think it's better to keep its name as `getManagedWallet$`.
> 
> What do you think?

The endpoint will last until we decided to revise the wallet endpoints and make breaking changes in the future, and we don't have plan to do it yet. In short term, the likeness of deprecating `/wallets/num-wallet/` is lower than the likeness to deprecate/rename `/wallets/managed-wallet/`.

shc261392 (approved):

PR #1399, opened 2022-03-22
feat: add network action order history detail page and loading indica…

ruizehung:
- Add network action order history detail page.
- Add loading indicator at network action history page.
- Rename `TransactionPage` to `ActivitiesPage`
- Extract code that fetch order history from bubble db to `order-history.service.ts`

## Test Plan
Demo video: https://youtu.be/fpX-fkR4gWA

If the result url is available.
image

```bash
➜  capture-lite git:(feature-add-network-action-order-detail-page) ✗ npm run test.ci             

> capture-lite@0.50.1 test.ci
> npm run preconfig && ng test --no-watch --no-progress --source-map=false --browsers=ChromeHeadlessCI


> capture-lite@0.50.1 preconfig
> node set-secret.js

A secret file has generated successfully at ./src/app/shared/dia-backend/secret.ts 

22 03 2022 13:56:42.981:INFO [karma-server]: Karma v6.3.4 server started at http://localhost:9876/
22 03 2022 13:56:42.982:INFO [launcher]: Launching browsers ChromeHeadlessCI with concurrency unlimited
22 03 2022 13:56:42.985:INFO [launcher]: Starting browser ChromeHeadless
22 03 2022 13:56:50.104:INFO [Chrome Headless 99.0.4844.74 (Mac OS 10.15.7)]: Connected on socket 7XiSMNorufmUSudAAAAB with id 54974266
ERROR: '<ion-refresher> must be used inside an <ion-content>'
Chrome Headless 99.0.4844.74 (Mac OS 10.15.7) MediaStore should delete atomicly FAILED
        Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
            at <Jasmine>
        Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
            at <Jasmine>
ERROR: 'NG0304: 'app-capture-transactions' is not a known element:
1. If 'app-capture-transactions' is an Angular component, then verify that it is part of this module.
2. If 'app-capture-transactions' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.'
Chrome Headless 99.0.4844.74 (Mac OS 10.15.7): Executed 221 of 221 (1 FAILED) (26.969 secs / 26.778 secs)
TOTAL: 1 FAILED, 220 SUCCESS

=============================== Coverage summary ===============================
Statements   : 51.35% ( 1849/3601 )
Branches     : 28.92% ( 295/1020 )
Functions    : 40.31% ( 643/1595 )
Lines        : 53.33% ( 1674/3139 )
================================================================================
➜  capture-lite git:(feature-add-network-action-order-detail-page) ✗ 
```


src/app/shared/actions/service/order-history.service.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  createOrderHistory$(networkAppOrder: NetworkAppOrder, cid: string) {
+    return defer(() =>
+      this.webCryptoApiSignatureProvider.publicKey$.pipe(
+        concatMap(publicKey =>
+          this.httpClient.post<BubbleCreateNewThingResponse>(
+            `${BUBBLE_DB_URL}/api/1.1/obj/order`,
+            {
+              asset_id_text: cid,
+              gas_fee_number: Number(networkAppOrder.fee),
```
shc261392, line 59: Using JavaScript number could lose precision for price and gas fee. Also I'm not sure what is the field type and precision on Bubble DB, but we should be careful about it.

Since backend still stored the full precision value, I think it is acceptable to store these values as regular floating point types as long as we don't use them to do calculation that requires 100% precision. But I would strongly recommend writing down this pitfall in network action doc.

Or maybe we could consider storing these value as String. When needed, Bubble side or App could use the full precision string and process them with JavaScript libraries that support big decimals.

shc261392 (commented):
Please take a look at the code comment

src/app/shared/actions/service/order-history.service.ts
ruizehung: @shc261392  Bubble db stores network action order price and gas as type `number`.  Also, I tried manually type in the full precision backend return into bubble db, and it lost the precision as well. So I guess even if we could convert it to number without losing precision on app side, Bubble couldn't store it without losing precision. 

So, we either 
1. Keep it as it is now and not worry about having bubble order history db storing price/gas fee with full precision
2. Update the price/gas fee and other `number` fields to `text` to support storing full precision
I need input from @tammyyang 

Screen Shot 2022-03-22 at 4 47 57 PM

tammyyang: Anyway which can lead us to "consistent and clear results" everywhere is a good method to me.

ruizehung: Seems like we couldn't update the field type after the field has been created, I'd say let's go with 1 and query backend using `order_id` if we need to do calculation with full precision.

ruizehung: Updated network action dev doc here: https://github.com/numbersprotocol/numbers-dev-doc/blob/main/docs/network_app/develop.md

shc261392 (approved):

PR #1506, opened 2022-04-04
feat: add icon to indicate capture with caption

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1202071964488862) by [Unito](https://www.unito.io)
┆Created By: Tammy Yang

### DIFF
--- a/src/app/features/home/capture-tab/capture-item/capture-item.component.html
+++ b/src/app/features/home/capture-tab/capture-item/capture-item.component.html
@@ -6,17 +6,26 @@
   name="hourglass-outline"
   class="collecting"
 ></ion-icon>
-<ion-icon
-  *ngIf="hasUploaded$ | ngrxPush"
-  src="/assets/images/media-id.svg"
-  class="uploaded"
->
-</ion-icon>
-<ion-icon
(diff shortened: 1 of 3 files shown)


sultanmyrza:
Current implementation looks like this ([demo](https://imgur.com/2VR4XN8)).
The current, implementation has 2 Issues!

1. (Not reactive) With the current implementation users always need to refresh the page to see if the caption icon is updated.
2. (Not optimal) for every item I call `diaBackendAssetRepository.fetchByProof$` to get `DiaBackendAsset` and then I get the caption aka  [`map(asset => asset.caption !== '')`](https://github.com/sultanmyrza/capture-lite/commit/c4dd34fa5e86cb74e5439a97c8a7bf3b9008bec8#diff-6bd903acaabfbf88b0091f6a4a56310eb096633a724d6ab8e4ea47ffc2963314R76)


**Is there a better way to get a caption for capture?**

shc261392 (approved):
I think the current approach is acceptable so I'll accept this pull request. We can optimize it further in the future.

PR #1540, opened 2022-04-22
Build bump to 0.54.2

sultanmyrza:


shc261392 (approved):
Looks good to me.

shc261392:
Discussion for `preview-video` plugin
https://dt42-numbers.slack.com/archives/C0323488MEJ/p1650612388174209
https://dt42-numbers.slack.com/archives/C0323488MEJ/p1650614603596759

[capture-lite]
We can choose from either
1. Disable the component test in `tsconfig.spec.json` by adding `"exclude": ["src/<path-to-the-file>/go-pro-media-viewer-with-native-player.component.ts"]`
2. Inside `GoProMediaViewerWithNativePlayerComponent`, check platform before calling any functions from `preview-video` plugin. If it is web, don’t call any unimplemented functions from `preview-video`.

[preview-video]
The previous change can be reverted.

PR #1603, opened 2022-05-11
fix(global.scss): default color for iOS safe areas

sultanmyrza-numbersprotocol:
safe area before after


shc261392 (approved):
Looks good to me

PR #1782, opened 2022-07-04
Feature qr code for deposit

sultanmyrza:
This is the result of pull request

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1202545955860671) by [Unito](https://www.unito.io)


shc261392:
@sultanmyrza I would suggest having a clear reminder to tell users they have to use BEP20 (or BSC, BNB Chain). Many wallet app have such hint because users may accidentally deposit tokens from wrong network to the wallet.

shc261392 (approved):
Looks good to me

PR #1792, opened 2022-07-05
Feature asset menu regardless of backend

sultanmyrza:
For now I just remove on item from options menu. After discussing with tammy we come to decision to remove "View Supporting Video On Ipfs" from menu options


sultanmyrza:
Ne need to add to 0.60.0 release. pull request is deprecated

PR #1805, opened 2022-07-12
add pre-view page before uploading capture

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1202592614295290) by [Unito](https://www.unito.io)

### DIFF
--- a/src/app/features/home/custom-camera/custom-camera.page.ts
+++ b/src/app/features/home/custom-camera/custom-camera.page.ts
@@ -2,9 +2,10 @@
 import { Location } from '@angular/common';
 import { Component, OnDestroy, OnInit } from '@angular/core';
 import { Router } from '@angular/router';
-import { PluginListenerHandle } from '@capacitor/core';
+import { Capacitor, PluginListenerHandle } from '@capacitor/core';
 import { UntilDestroy } from '@ngneat/until-destroy';
 import { CaptureResult, PreviewCamera } from '@numbersprotocol/preview-camera';
+import { BehaviorSubject } from 'rxjs';
 import { ErrorService } from '../../../shared/error/error.service';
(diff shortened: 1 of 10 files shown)


src/app/features/home/custom-camera/custom-camera.page.ts
```diff
@@ -142,6 +178,13 @@ export class CustomCameraPage implements OnInit, OnDestroy {
     });
   }
 
+  private removeCurrentCapture() {
+    // TODO: remove file this.curCaptureFilePath to free space
```
bafu, line 182: @sultanmyrza Can we remove the unconfirmed file in this PR, or can you share some details of the difficulty, and we can open a following-up task for it?

sultanmyrza: @bafu, when user take photos or videos it's kept in file system so we need to clean up files. I don't think it's hard by I will add another pull req that will implement this feature (today)

sultanmyrza:
Here is the result for "[fix(pre-publish.page): add header, footer to match UX](https://github.com/numbersprotocol/capture-lite/pull/1805/commits/24ecfc4576eb624ea6934cc66fb3ea35710fb2f7)"

https://imgur.com/a/84zBQPM (iOS and Android)

bafu (approved):
Accept this PR with 3 known issues

1. The positions of the back and confirm buttons, and the header beight are not ideal on my phone. Considering that the phone resolution is not general, I will accept this PR first.
1. No black footer behind the camera button.
1. Unconfirmed temporary photos will not be dropped.

[Details](https://app.asana.com/0/0/1202546228465323/1200919144392346/f)

PR #1826, opened 2022-07-19
Camera rebrand feature flash

sultanmyrza:


src/app/features/home/custom-camera/custom-camera.service.ts
```diff
@@ -98,6 +100,23 @@ export class CustomCameraService {
...
+    if (this.isNativePlatform) {
+      return await PreviewCamera.isTorchOn();
+    }
+    return { result: false };
+  }
+
+  async enableTorch(enable: boolean): Promise<void> {
+    if (this.isNativePlatform) {
+      return await PreviewCamera.enableTorch({ enable });
+    }
```
bafu, line 113: @sultanmyrza Should we handle the "else" case?

sultanmyrza-numbersprotocol: @bafu we do not support web at this point so I don't think we need `else` block.

bafu: The return type is Promise, and the function will return implicitly in the "else" case.

In `camera.enableTorch` (L183), it just awaits the  `service.enableTorch`, so functionality-wise it should work. It might increase future readability for the other collaborators to indicate the return value in the "else" case.

bafu (approved):

PR #1849, opened 2022-07-25
Feature appsflyer sdk try2

sultanmyrza:


bafu (approved):

bafu:
Note: `trackUserOpenedWalletsPage` in `apps-flyer.service.ts` is not used currently. We keep it because it's useful as a reference for how to send a log event to AppsFlyer.

PR #1919, opened 2022-08-08
fix ios status bar color, set to white

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1202765748989243) by [Unito](https://www.unito.io)


bafu:
@sultanmyrza 

1. Will this fix work for both light/dark modes?
2. The space indents are different from the current tab indents. Please help replace them with space indents for consistency.

bafu:
Got confirmation from Sam that the fix will work for both light/dark modes.

bafu (approved):

PR #1924, opened 2022-08-09
fix(details.page): iframe url

sultanmyrza:


bafu:
We do not use testing version URLs in Capture App. This PR is for fixing the issue.

PR #1958, opened 2022-08-23
change(explore-tab): iframe url to version-qa

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1202846242800227) by [Unito](https://www.unito.io)


bafu:
In the future, we will create two versions of Android/iOS Apps

```
# General mode
capture --- with Bubble qa-release (Android APK which can be installed locally)
        `-- with Bubble live (which can be installed via Android/iOS Stores)
```

When there is any huge update, and we need to iterate the App quickly

```
# Iterating mode
capture -- with Bubble qa-release (Android APK which can be installed locally)
        `-- with Bubble qa-release (which can be installed via Android/iOS Stores)
```

until the feature is stable. The project manager will decide when will be in the iterating mode.

bafu (approved):

PR #2241, opened 2022-10-30
Fix image editor keep black and white filter only

sultanmyrza:
**NOTE: Should be included in the v221101 release.**

Customize the editor to match capture rebrand UI/UX and remove default buttons, paddings, borders of Image editor plugin. Here is the [claap](https://app.claap.io/numbers-protocol/black-and-white-filter-only-c-O35CsUM4Uy-GH2Xal-gVYPM) demo.


sultanmyrza:
re-created same [PR](https://github.com/numbersprotocol/capture-lite/pull/2257) with clean commits.

PR #2263, opened 2022-11-01
fix GitHub build-ios-prod job

sultanmyrza:
Currently, the **build-ios-prod job** is failing with error:

```
error: exportArchive: Provisioning profile "NumbersAppDistributionV2" doesn't include signing certificate "iPhone Distribution: Numbersprotocol Inc. (G7NB5YCKAP)".
```

I found 1 place where we still have `NumbersAppDistributionV2` and change it to `NumbersAppDistributionV4`.


sultanmyrza:
checked out from old dev

PR #2329, opened 2022-11-18
Feat upgrade capacitor from 3 to 4 try 1

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1203400014825176) by [Unito](https://www.unito.io)


src/app/shared/capacitor-plugins/mock-app-plugin.ts
```diff
@@ -20,10 +20,47 @@ export class MockAppPlugin implements AppPlugin {
     throw new Error('Method not implemented.');
   }
 
+  // addListener(
```
bafu, line 23: @sultanmyrza Suggest adding a comment about why commenting out and keep these codes so that we can catch up on the reason in the future quickly.

bafu (changes requested):

src/app/shared/capacitor-plugins/mock-app-plugin.ts
```diff
@@ -20,35 +20,6 @@ export class MockAppPlugin implements AppPlugin {
     throw new Error('Method not implemented.');
   }
 
-  // addListener(
```
sultanmyrza: In the previous commit, I forgot to remove this commented code. It's safe to remove these lines.

bafu (approved):

PR #2372, opened 2022-12-04
add missing files generated after `npx cap sync`

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1203486465100300) by [Unito](https://www.unito.io)

### DIFF
--- a/android/app/src/main/assets/capacitor.config.json
+++ b/android/app/src/main/assets/capacitor.config.json
@@ -21,5 +21,8 @@
 			"iconColor": "#564dfc"
 		}
 	},
-	"cordova": {}
+	"cordova": {},
+	"server": {
+		"androidScheme": "https"
+	}
 }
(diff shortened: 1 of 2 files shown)


bafu (approved):

bafu:
@sultanmyrza `androidScheme` had been added into `<capture-root>/capacitor.config.json` in commit `f2ce6bd`.

What is the reason that `npx cap sync` does not update `android/app/src/main/assets/capacitor.config.json`?

sultanmyrza:
> @sultanmyrza `androidScheme` had been added into `<capture-root>/capacitor.config.json` in commit `f2ce6bd`.
> 
> What is the reason that `npx cap sync` does not update `android/app/src/main/assets/capacitor.config.json`?

@bafu , yes it was added in `f2ce6bd` but after adding it to the capacitor config in the root directory I forget to run npx cap sync and committed.

But later after running npx cap syn it added androidScheme to anroid/../capacitor.config.json automatically 

So usually when adding things to root/capacitor config we need to run `npx cap sync` but I forgot that's why we have this PR.

bafu (approved):

PR #2457, opened 2023-01-05
change: disable force webview darkmode on android

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1203655259332063) by [Unito](https://www.unito.io)


bafu:
The white flash workaround causes many UI side effects, and we decide to revert the workaround.

To revert the workaround, we will apply PR 2457 and PR 2460.

bafu (approved):

PR #2466, opened 2023-01-09
fix(transfer.page): set QR code text to white

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1203672902353605) by [Unito](https://www.unito.io)

### DIFF
--- a/src/app/features/wallets/transfer/transfer.page.scss
+++ b/src/app/features/wallets/transfer/transfer.page.scss
@@ -182,6 +182,7 @@ ion-card {
   display: flex;
   flex-direction: column;
   align-items: center;
+  color: white !important; /* stylelint-disable-line declaration-no-important */
 }
 
 .qr-code-info {


tammyyang:
@sultanmyrza this is under the assumption that we do not allow users to switch between dark mode <> light mode, right? I assume if the light mode is enabled, this setting may cause some issues.

sultanmyrza:
> @sultanmyrza this is under the assumption that we do not allow users to switch between dark mode <> light mode, right? I assume if the light mode is enabled, this setting may cause some issues.

@tammyyang, correct. This PR is not considering light mode at all.

tammyyang:
@sultanmyrza do you think if we should add something like this for light mode now to prevent the issue from happening or just deal with it later :P?

```
.asset-wallet-qr-code-container {
  max-width: 300px;
  margin: 0 auto;
  display: flex;
  flex-direction: column;
  align-items: center;
  color: white !important; /* stylelint-disable-line declaration-no-important */
}

@media (prefers-color-scheme: light) {
  .asset-wallet-qr-code-container {
    color: black;
  }
}
```

sultanmyrza:
> @sultanmyrza do you think if we should add something like this for light mode now to prevent the issue from happening or just deal with it later :P?
> 
> ```
> .asset-wallet-qr-code-container {
>   max-width: 300px;
>   margin: 0 auto;
>   display: flex;
>   flex-direction: column;
>   align-items: center;
>   color: white !important; /* stylelint-disable-line declaration-no-important */
> }
> 
> @media (prefers-color-scheme: light) {
>   .asset-wallet-qr-code-container {
>     color: black;
>   }
> }
> ```

@tammyyang , ok let me try and test it on light/dark modes again.

tammyyang:
@sultanmyrza allow me to clarify. 

1. When you set dark mode "off", the QR code needs to be "white" because the background is black even if the dark mode is off.
2. When you set the dark mode "on", the QR code still needs to be "white" because the background is black (because the dark mode is on now). Plus, the color setting in the "prefer-color-theme: light" is triggered (which it should not, because the dark mode is "on" now.

Are these two the main problems?

Is my understanding correct?

sultanmyrza:
@tammyyang ,  yes "prefer-color-theme: light" should not be triggered at all because we [set dark mode](https://github.com/numbersprotocol/capture-lite/blob/732dcc56da7d8737d60ef271aee2405a141394c4/src/app/app.component.ts#L67) (not WebView but css class) at app launch.

sultanmyrza:
@tammyyang, at this point no matter user sets system dark/light mode our app always show dark UI elements, and that QR code text always should be white, unless we want to add light theme in the future.

tammyyang:
> unless we want to add light theme in the future.

Yes, I think we will want to add light theme in the future. That's why I am a little bit concerned about this. 
Can we try one last thing? 
Screenshot 2023-01-09 205514

According to the [Ionic document](https://ionicframework.com/docs/theming/dark-mode#css-class-fallback), we should have `:root` and the fallback function. Let's give it a try?

sultanmyrza:
@tammyyang, For now we can conditionally accept this pr as tech debit, and later fix it by this future [asana](https://app.asana.com/0/1201016280880500/1203685565125839/f) card (details explained in the card).

tammyyang (approved):
Due to t[his comment](https://github.com/numbersprotocol/capture-lite/pull/2466#issuecomment-1376755473), a task is open for the future development: https://app.asana.com/0/1201016280880500/1203685565125839/f

PR #2498, opened 2023-01-19
Feat v230112 rebranded app menu

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1203781772835372) by [Unito](https://www.unito.io)

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -82,6 +82,13 @@ const routes: Routes = [
         m => m.InvitationPageModule
       ),
   },
+  {
+    path: 'data-policy',
+    loadChildren: () =>
+      import('./features/data-policy/data-policy.module').then(
+        m => m.DataPolicyPageModule
+      ),
+  },
 ];
 @NgModule({
   imports: [
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -89,6 +89,13 @@ const routes: Routes = [
         m => m.DataPolicyPageModule
       ),
   },
+  {
+    path: 'terms-of-use',
+    loadChildren: () =>
+      import('./features/terms-of-use/terms-of-use.module').then(
+        m => m.TermsOfUsePageModule
+      ),
+  },
 ];
 @NgModule({
   imports: [
--- a/src/app/features/data-policy/data-policy.page.ts
+++ b/src/app/features/data-policy/data-policy.page.ts
@@ -1,0 +1,45 @@
+import { Component } from '@angular/core';
+import { NavController } from '@ionic/angular';
+import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
+import { BehaviorSubject, fromEvent } from 'rxjs';
+import { tap } from 'rxjs/operators';
+import { BUBBLE_IFRAME_URL } from '../../shared/dia-backend/secret';
+import { NetworkService } from '../../shared/network/network.service';
+
+@UntilDestroy()
+@Component({
+  selector: 'app-data-policy',
+  templateUrl: './data-policy.page.html',
+  styleUrls: ['./data-policy.page.scss'],
+})
+export class DataPolicyPage {
+  readonly networkConnected$ = this.networkService.connected$;
+  readonly dataPolicyUrl = `${BUBBLE_IFRAME_URL}/data_policy`;
+  readonly iframeLoaded$ = new BehaviorSubject(false);
+
+  constructor(
+    private readonly networkService: NetworkService,
+    private readonly navController: NavController
+  ) {}
+
+  ionViewDidEnter() {
+    this.processIframeEvents();
+  }
+
+  processIframeEvents() {
+    fromEvent(window, 'message')
+      .pipe(
+        tap(event => {
+          const postMessageEvent = event as MessageEvent;
+          if (postMessageEvent.data === 'iframe-on-load') {
+            this.iframeLoaded$.next(true);
+          }
+          if (postMessageEvent.data === 'iframeBackButtonClicked') {
+            this.navController.back();
+          }
+        }),
+        untilDestroyed(this)
+      )
+      .subscribe();
+  }
+}
--- a/src/app/features/home/home.page.ts
+++ b/src/app/features/home/home.page.ts
@@ -14,6 +14,7 @@ import { combineLatest, defer, EMPTY, iif, of } from 'rxjs';
 import {
   catchError,
   concatMap,
+  concatMapTo,
   first,
   map,
   startWith,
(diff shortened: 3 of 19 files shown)


src/app/features/data-policy/data-policy.page.ts
```diff
@@ -0,0 +1,45 @@
...
+          }
+          if (postMessageEvent.data === 'iframeBackButtonClicked') {
+            this.navController.back();
+          }
+        }),
+        untilDestroyed(this)
+      )
+      .subscribe();
+  }
+}
```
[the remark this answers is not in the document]
sultanmyrza: @tammyyang regarding improvement suggestions
- If we use `@HostListener` then it would look like `@HostListener('window:message', ['$event']) onMessage(event: Event) {
    console.log('Process the message!');
  }` to me using `fromEvent` (line: 30) is more clear and has some auto-completion. So I would ignore this suggestion.
- it's already marked as `readonly`, I would ignore this suggestion.
- a good point need to do this for all iframes inside the capture app 
  - [[Code Quality + Security] capture-app-ionic. Validate the origin of the message event (iframe postMessage), this is to ensure that the event is sent from a trusted source.](https://app.asana.com/0/1201016280880500/1203794276110740/f)
  - [[Code Quality + Security] capture-app-iframe. Validate the origin of the message event (iframe postMessage), this is to ensure that the event is sent from a trusted source.](https://app.asana.com/0/0/1203794276110743/f)
- Does not make sense.

src/app/app-routing.module.ts
```diff
@@ -89,6 +89,13 @@ const routes: Routes = [
         m => m.DataPolicyPageModule
       ),
   },
+  {
+    path: 'terms-of-use',
+    loadChildren: () =>
+      import('./features/terms-of-use/terms-of-use.module').then(
+        m => m.TermsOfUsePageModule
+      ),
+  },
```
sultanmyrza, line 98: No need for ChatGPT, just adding routing standard code.

src/app/features/terms-of-use/terms-of-use-routing.module.ts
```diff
@@ -0,0 +1,17 @@
...
+    path: '',
+    component: TermsOfUsePage,
+  },
+];
+
+@NgModule({
+  imports: [RouterModule.forChild(routes)],
+  exports: [RouterModule],
+})
+export class TermsOfUsePageRoutingModule {}
```
sultanmyrza, line 17: No need for ChatGPT. Standart code generated by Angular CLI.

src/app/features/terms-of-use/terms-of-use.module.ts
```diff
@@ -0,0 +1,12 @@
...
+import { TermsOfUsePageRoutingModule } from './terms-of-use-routing.module';
+
+import { SharedModule } from '../../shared/shared.module';
+import { TermsOfUsePage } from './terms-of-use.page';
+
+@NgModule({
+  imports: [SharedModule, TermsOfUsePageRoutingModule],
+  declarations: [TermsOfUsePage],
+})
+export class TermsOfUsePageModule {}
```
sultanmyrza, line 12: No need for ChatGPT. Standart code generated by Angular CLI.

src/app/features/terms-of-use/terms-of-use.page.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+      fixture = TestBed.createComponent(TermsOfUsePage);
+      component = fixture.componentInstance;
+      fixture.detectChanges();
+    })
+  );
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
sultanmyrza, line 26: No need for ChatGPT. Standart code generated by Angular CLI.

src/app/features/terms-of-use/terms-of-use.page.ts
```diff
@@ -0,0 +1,45 @@
...
+          }
+          if (postMessageEvent.data === 'iframeBackButtonClicked') {
+            this.navController.back();
+          }
+        }),
+        untilDestroyed(this)
+      )
+      .subscribe();
+  }
+}
```
sultanmyrza, line 45: `ChatGPT code review for git diff 6c6d81d3..45109709 "*terms-of-use.page.ts"`

This code defines a `TermsOfUsePage` component for an Ionic application. The component displays the terms of use in an iframe, and handles events coming from the iframe. It listens to the network connection status and iframe's load event by using observables, and it also listens to the iframe's back button event by using event listener.

Improvement suggestions:

-   Extract the common functionality of the two pages `DataPolicyPage` and `TermsOfUsePage` into a shared service or a base page class to avoid duplicated code
-   Instead of using a magic string "iframe-on-load", you can define a constant variable for this string value, this can make your code more readable and maintainable
-   The `iframeLoaded$` variable should be defined as a private variable and not exposed as a public variable.
-   Use `const` keyword instead of `let` for variables that are not reassigned.
-   Make sure that the URL used to load the iframe is a trusted source, as using untrusted URLs can lead to security vulnerabilities such as XSS attacks.

sultanmyrza: @tammyyang regarding improvement suggestions:

- Make sense, but it's better when things are repeated more than 2 times otherwise it's the same amount of code. Would ignore this suggestion.
- Make sense, related asana cards are created.
  - [[Code Quality + Security] capture-app-ionic. Unify iframe postMessage format and extract to ENUMs](https://app.asana.com/0/1201016280880500/1203794276110745/f)
  - [[Code Quality + Security] capture-app-iframe. Unify iframe postMessage format](https://app.asana.com/0/0/1203794276110749/f)
- Agree but maybe we can keep it public.
- For this commit for variables that are not reassigned I use `const` and `readonly` already.
- iframe URL is assigned at the CI build phase. Therefore URL for the iframe is highly likely to be safe

src/app/features/home/home.page.ts
```diff
@@ -370,4 +380,25 @@ export class HomePage {
...
+    )
+      .pipe(
+        concatMap(result =>
+          iif(() => result, this.blockingActionService.run$(action$))
+        ),
+        untilDestroyed(this)
+      )
+      .subscribe();
+  }
 }
```
sultanmyrza, line 404: `ChatGPT code review for git diff 45109709..119d330d "*home.page.ts"`

-   In the `logout` function, it is recommended to use the `concatMap` operator instead of chaining multiple `concatMapTo` operators. This way, the logout process will be executed sequentially and not concurrently.
-   In the `logout` function, it is recommended to use the `switchMap` operator instead of `concatMap` as it ensures that only the latest inner observable is subscribed to, so it's not possible to have multiple logout process at the same time and also it's more efficient because it unsubscribes from the previous inner observable.
-   It is recommended to call `blockingActionService.block()` before starting the logout process and call `blockingActionService.unblock()` after the logout process has finished. This way, the user is informed that the logout process is in progress and can't perform any other action in the meanwhile.
-   It is recommended to import `reloadApp` from a more specific location, rather than from the root of the project.
-   It is recommended to move the `downloadExpiredPostCaptures` function call to the constructor of the service that's responsible for handling capture logic, rather than calling it in the `HomePage` constructor.
-   It is recommended to use `onDidDismiss` observable of the modal to unblock the blocking action instead of using a setTimeout.

src/app/features/terms-of-use/terms-of-use.page.ts
tammyyang: Arrange suggestion tasks in the next sprint

tammyyang (commented):
Two tasks for improving the code quality are arranged in the next sprint.

PR #2502, opened 2023-01-19
fix(details.page): prevent collected captures to edit, unpublish

sultanmyrza:


tammyyang:
`ChatGPT code review for src/app/features/home/details/details.page.ts`

This code seems to be adding a new property to the class called isCollectedCapture$, which is an observable that maps the type$ observable to a boolean that indicates whether the type is equal to 'post-capture'.

Then, it modifies the openOptionsMenuEvenOffline method to use the isCollectedCapture$ observable to conditionally show the "Edit" and "Unpublish" options in the action sheet.

Overall, this change seems to be making the options in the action sheet conditional on whether the capture is a "post-capture" or not, which is a good idea.

A possible improvement that can be done is that, the method can be refactor to use the async pipe and handle the promise resolution inside the subscribe block.

Also, it is a good practice to use the take(1) instead of first() as it is a rxjs operator that will complete the observable after the first emission (same as first).

PR #2529, opened 2023-01-29
Feat v230131 lazy login

sultanmyrza:


src/app/features/home/explore-tab/explore-tab/explore-tab.component.ts
```diff
@@ -19,10 +19,12 @@ export class ExploreTabComponent {
...
-      const url = `${BUBBLE_IFRAME_URL}/?token=${token.access}&refresh_token=${token.refresh}`;
-      return this.sanitizer.bypassSecurityTrustResourceUrl(url);
-    })
+      if (!token) return BUBBLE_IFRAME_URL;
+      return `${BUBBLE_IFRAME_URL}/?token=${token.access}&refresh_token=${token.refresh}`;
+    }),
+    map(url => this.sanitizer.bypassSecurityTrustResourceUrl(url))
   );
 
   readonly networkConnected$ = this.networkService.connected$;
```
sultanmyrza, line 30: **ChatGPT response for `git diff 6df9e825..4b49d31f "*explore-tab.component.ts"`**
Here are some improvement suggestions for the code:

-   Consider renaming the component file to follow a consistent naming convention. For example, you could use "explore-tab.component.ts".
-   Move the import statements for third-party libraries to the top, followed by internal imports.
-   Use `startWith` operator to provide an initial value for the stream, so it emits a value on subscription even if the source stream is empty.
-   Use a ternary operator in the `map` operator to simplify the condition checking whether the token is available or not.
-   Instead of constructing the URL string in two separate steps, you could use string templates to construct the URL in one step.

Here's the revised code with these suggestions applied:



```typescript
import { Component, ElementRef, ViewChild } from '@angular/core';
import { DomSanitizer } from '@angular/platform-browser';
import { combineLatest, of } from 'rxjs';
import { map, startWith, tap } from 'rxjs/operators';

import { DiaBackendAuthService } from '../../../../shared/dia-backend/auth/dia-backend-auth.service';
import { BUBBLE_IFRAME_URL } from '../../../../shared/dia-backend/secret';
import { IframeService } from '../../../../shared/iframe/iframe.service';

@Component({
  selector: 'app-explore-tab',
  templateUrl: './explore-tab.component.html',
  styleUrls: ['./explore-tab.component.scss']
})
export class ExploreTabComponent {
  url$ = combineLatest([
    this.diaBackendAuthService.cachedQueryJWTToken$,
    this.iframeService.exploreTabRefreshRequested$,
  ]).pipe(
    startWith([undefined, undefined]),
    map(([token, _]) => `${BUBBLE_IFRAME_URL}/?token=${token?.access}&refresh_token=${token?.refresh}`),
    map(url => this.sanitizer.bypassSecurityTrustResourceUrl(url))
  );

  constructor(private sanitizer: DomSanitizer, private diaBackendAuthService: DiaBackendAuthService, private iframeService: IframeService) {}
}
```

sultanmyrza: @tammyyang, regarding improvements suggestions by ChatGPT:
- Make no sense.
- Make no sense.
- Already done.
- Using if make intention more clear
- Already good.

src/app/features/home/home.page.ts
```diff
@@ -377,10 +380,16 @@ export class HomePage {
...
     await this.userGuideService.setHasOpenedInboxTab(true);
   }
 
+  private async redirectToLoginPage() {
+    this.router.navigate(['/', 'login']);
+  }
+
   logout() {
     const action$ = defer(() => this.mediaStore.clear()).pipe(
       concatMapTo(defer(() => this.database.clear())),
```
sultanmyrza, line 395: **ChatGPT response for git diff 4b49d31f.. 8677a8b0 "*home.page.ts"**

Here are some improvement suggestions for the code:

-   Consider extracting the `async ionViewDidEnter()` logic into a separate method that returns an observable, and calling it from `ionViewDidEnter`
-   Extract the code for checking if the user is logged in into a separate method and reuse it where necessary
-   Remove the `async` keyword from the `captureWithCustomCamera` method, it is not necessary as it only calls other functions which are not async
-   Replace the `of(this.onboardingService.isNewLogin)` in `ionViewDidEnter` method with `of(await this.onboardingService.isNewLogin)` so that the value of `isNewLogin` is evaluated before it is passed to the `migrate$` method
-   Add a default case in the `switch` statement in the `presentCaptureActions$` method to handle unexpected values of `this.selectedTabIndex`
-   Use a `switchMap` operator instead of a `switchTapTo` operator in the `ionViewDidEnter` method, as `switchMap` would prevent the creation of multiple subscriptions when navigating to the same component multiple times.
-   Add an explicit type for the error parameter in the catchError operator to provide more clarity.

sultanmyrza: @tammyyang regarding improvements:
- Opinionated 
- Opinionated
- Still need async
- Existing solution is working as expected I tried before rxjs way and restore capture dialog was not showing so lets leave it as it is (aka working)
- `switchTapTo` is working as expected all this time so no need to change
- Overkill, errors can come from different places `migrate`, `promptAppUpdateIfAny`, `onboardingRedirect`, `showUserGuidesOnHomePage` so using generic type `unknow` make sense.

PR #2589, opened 2023-02-15
refactor: to enums + switch for iframe post message events

sultanmyrza:


src/app/features/data-policy/data-policy.page.ts
```diff
@@ -31,11 +32,14 @@ export class DataPolicyPage {
...
+          switch (data) {
+            case BubbleToIonicPostMessage.IFRAME_ON_LOAD:
+              this.iframeLoaded$.next(true);
+              break;
+            case BubbleToIonicPostMessage.IFRAME_BACK_BUTTON_CLICKED:
+              this.navController.back();
+              break;
           }
         }),
         untilDestroyed(this)
```
sultanmyrza, line 45: **ChatGPT response for**  `git show a9e4232dfd700d07b6a61ec73910927941b129ae src/app/features/data-policy/data-policy.page.ts`

1.  It's good to use enums instead of string literals to avoid typo errors.
2.  It's better to define the type of the message event data for better code clarity.
3.  Consider adding a default case to the switch statement for unexpected post message events.

sultanmyrza: @tammyyang regarding ChatGPT response:

1. ignore
2. make sense, but currently we just pass string as a message so I would ignore for now.
3. agree, I missed `default: case`. Created asana card [[Code Quality] Consider adding a default case to the switch statement for unexpected post message events](https://app.asana.com/0/1201016280880500/1204028658078831/f) _Find all switch statements in capture-app and make sure it has default case._

src/app/features/faq/faq.page.ts
```diff
@@ -31,11 +32,14 @@ export class FaqPage {
...
+          switch (data) {
+            case BubbleToIonicPostMessage.IFRAME_ON_LOAD:
+              this.iframeLoaded$.next(true);
+              break;
+            case BubbleToIonicPostMessage.IFRAME_BACK_BUTTON_CLICKED:
+              this.navController.back();
+              break;
           }
         }),
         untilDestroyed(this)
```
sultanmyrza, line 45: **ChatGPT response** for `git show a9e4232dfd700d07b6a61ec73910927941b129ae src/app/features/faq/faq.page.ts `

The code looks good. Since it is a similar change to the previous one, it might be good to extract the switch statement into a separate function that can be reused across different components to avoid duplicating code.

sultanmyrza: @tammyyang regarding ChatGPT response :

No need to extract to separate methods since all of them one liners

src/app/features/home/details/edit-caption/edit-caption.page.ts
```diff
@@ -52,14 +53,15 @@ export class EditCaptionPage {
...
+            case BubbleToIonicPostMessage.EDIT_CAPTION_CANCEL:
+              this.navController.back();
+              break;
+            case BubbleToIonicPostMessage.EDIT_CAPTION_SAVE:
+              this.iframeService.refreshDetailsPageIframe();
+              this.navController.back();
+              break;
           }
         }),
         untilDestroyed(this)
```
sultanmyrza, line 67: **ChatGPT reponse** for ` git show a9e4232dfd700d07b6a61ec73910927941b129ae src/app/features/home/details/edit-caption/edit-caption.page.ts ` 

It looks like a good code refactor that uses enums and switch statements for post message events. It could be further improved by adding error handling in case the received data is not of the expected type, and by using stricter type checking for the BubbleToIonicPostMessage enum, such as adding a union type of string literals for its values.

sultanmyrza: @tammyyang:

For postMessages we don't need to throw error if we got message that we are not interested we just need to skip it so throwing error does not make sense just reacting to events that we are interested is what we need.

src/app/features/home/explore-tab/explore-tab/explore-tab.component.ts
```diff
@@ -45,7 +46,7 @@ export class ExploreTabComponent {
 
   navigateBackExploreIframe() {
     this.exploreIframe?.nativeElement.contentWindow?.postMessage(
-      'android-back-button',
+      IonicToBubblePostMessage.ANDROID_BACK_BUTTON,
       BUBBLE_IFRAME_URL
     );
   }
```
sultanmyrza, line 52: Too small for ChatGPT

src/app/features/terms-of-use/terms-of-use.page.ts
```diff
@@ -31,11 +32,14 @@ export class TermsOfUsePage {
...
+          switch (data) {
+            case BubbleToIonicPostMessage.IFRAME_ON_LOAD:
+              this.iframeLoaded$.next(true);
+              break;
+            case BubbleToIonicPostMessage.IFRAME_BACK_BUTTON_CLICKED:
+              this.navController.back();
+              break;
           }
         }),
         untilDestroyed(this)
```
sultanmyrza, line 45: The code changes seem fine. Here are a few suggestions for improvement:
**ChatGPT response for** `git show a9e4232dfd700d07b6a61ec73910927941b129ae src/app/features/terms-of-use/terms-of-use.page.ts `

1.  Use more descriptive variable names instead of generic names like `data` or `event`.
2.  Consider using a default case in the switch statement to handle unexpected values.
3.  If `BubbleToIonicPostMessage` is only used in this file, consider importing it with a relative path instead of a full path.

sultanmyrza: @tammyyang:
1. opinionated
2. already created asana task (see [comment](https://github.com/numbersprotocol/capture-lite/pull/2589/commits/a9e4232dfd700d07b6a61ec73910927941b129ae#r1115325035))
3. opinionated

src/app/shared/iframe/iframe.ts
```diff
@@ -0,0 +1,10 @@
+export enum BubbleToIonicPostMessage {
+  IFRAME_ON_LOAD = 'iframe-on-load',
+  IFRAME_BACK_BUTTON_CLICKED = 'iframe-back-button-clicked',
+  EDIT_CAPTION_CANCEL = 'edit-caption-cancel',
+  EDIT_CAPTION_SAVE = 'edit-caption-save',
+}
+
+export enum IonicToBubblePostMessage {
+  ANDROID_BACK_BUTTON = 'android-back-button',
+}
```
sultanmyrza, line 10: **by ChatGPT**
The code seems to be a refactoring of the way post message events are handled in iframes, with the introduction of enums and switch statements. Here are some suggestions:

-   It is good practice to have clear and concise commit messages. Instead of "refactor: to enums + switch for iframe post message events," consider something like "Refactor iframe post message event handling with enums and switch statements."
-   In the `BubbleToIonicPostMessage` enum, the values should be in uppercase letters, separated by underscores, to follow common naming conventions for enums. For example, `IFRAME_ON_LOAD` instead of `iframe-on-load`.
-   Consider adding comments to explain the purpose of the enums and the values they contain, especially if they are used across multiple files and/or projects.
-   In the `EditCaptionPage` and `TermsOfUsePage`, the `postMessageEvent.data` is being cast to the `BubbleToIonicPostMessage` type without any error handling in case the value is not a member of the enum. It may be a good idea to add a default case to the switch statement that logs an error message and/or throws an error if the value is not recognized.
-   Consider using a more descriptive name for the enums, especially if they are used in multiple contexts. For example, instead of `BubbleToIonicPostMessage`, consider something like `IframeEvent`.

sultanmyrza: @tammyyang :
- opinionated
- opinionated although good suggestion but it will require changes from bubble side as well so would keep lower case + underscore
- asana [card](https://app.asana.com/0/1201016280880500/1204028658078827/f) created.
- opinionated

PR #2590, opened 2023-02-15
feat(home.page): swap explore tab with profile tab

sultanmyrza:


src/app/features/home/home.page.ts
```diff
@@ -363,14 +362,17 @@ export class HomePage {
...
     }
   }
 
   // eslint-disable-next-line class-methods-use-this
   navigateToExploreTab() {
-    if (this.selectedTabIndex === 0) {
+    if (this.selectedTabIndex === this.exploreTabIndex) {
       // window.location.reload();
       this.iframeService.refreshExploreTabIframe();
     }
```
sultanmyrza, line 378: **ChatGPT response** 
1.  Use more descriptive variable names instead of comments, e.g. `const initialTabIndex = 0` instead of `private readonly initialTabIndex = 0`.
2.  Use consistent naming convention for variables, e.g. `exploreTabIndex` should be `exploreIndex`.
3.  Instead of hardcoding `this.afterCaptureTabIndex`, use the constant `this.initialTabIndex`.
4.  In `capture()` and `ionViewWillEnter()`, the line `this.selectedTabIndex = this.afterCaptureTabIndex` is duplicated. Consider creating a separate function to set the selected tab index to avoid duplication.
5.  Consider using an enum for tab indexes instead of hardcoding them in the code.
6.  Add comments to explain the purpose of `shouldNavigateBackExploreIframe()` and `navigateToExploreTab()`.

sultanmyrza: @tammyyang regarding ChatGPT response:
1. `initialTabIndex` is private and read only so using `const` make less sense.
2. `exploreTabIndex` make more sense.
3. `afterCaptureTabIndex` is tab that should be shown after user took photo/video no need to change.
4.  make sense but I would recomment to skip this.
5. I would skip.
6. I believe `shouldNavigateBackExploreIframe(`) and `navigateToExploreTab()` are self explanatory

tammyyang: Add this as follow up task https://app.asana.com/0/1201016280880500/1204094724972082/f

tammyyang (commented):
Review done

tammyyang:
This one is also relevant https://app.asana.com/0/1201016280880500/1204094724972101/f

PR #2614, opened 2023-02-22
fix(wallet.page): unify iframe post messages

sultanmyrza:


src/app/features/wallets/wallets.page.ts
```diff
@@ -56,23 +57,24 @@ export class WalletsPage {
...
               break;
-            case 'iframe-copy-to-clipboard-integrity-wallet':
+            case BubbleToIonicPostMessage.IFRAME_COPY_TO_CLIPBOARD_INTEGRITY_WALLET:
               this.copyToClipboardIntegrityWallet();
               break;
-            case 'iframe-copy-to-clipboard-private-key':
+            case BubbleToIonicPostMessage.IFRAME_COPY_TO_CLIPBOARD_PRIVATE_KEY:
               this.copyToClipboardPrivateKey();
               break;
             default:
```
sultanmyrza, line 80: TIP: to get diff for this file I use this command `git show 5cef4487d6d99ec3487fed32a6a976b2dbfe3ab4 "*wallets.page.ts" | pbcopy` on macOS `pbcopy` will copy result directly to clipboard. Then I just use this pattern 

> Please code review with improvement suggestions for the following: 
> < RESLUT OF GIT SHOW >

**ChatGPT response**
It's difficult to provide improvement suggestions without seeing the implementation of the `BubbleToIonicPostMessage` enum, but here are some general suggestions for this code:

1.   Add comments explaining what each case does.
1.   Consider extracting the logic for handling each case into separate methods to improve readability and maintainability.
1.   Consider using a switch/case statement with a default case that throws an error if an unknown message is received, to catch unexpected scenarios and prevent the app from breaking silently.

sultanmyrza: @tammyyang, regarding ChatGPT suggestions:
1. Cases are selfexplanatory no need for comments.
2. Already extracted to one line methods no need to change
3. In our case if we get postMessage that we are not interested then we just silently skip it, no need to throw error msg etc.

SUMMARY: As a result no need to create follow up asana card for code improvements etc.

src/app/shared/iframe/iframe.ts
```diff
@@ -3,6 +3,10 @@ export enum BubbleToIonicPostMessage {
   IFRAME_BACK_BUTTON_CLICKED = 'iframe-back-button-clicked',
   EDIT_CAPTION_CANCEL = 'edit-caption-cancel',
   EDIT_CAPTION_SAVE = 'edit-caption-save',
+  IFRAME_BUY_NUM_BUTTON_CLICKED = 'iframe-buy-num-button-clicked',
+  IFRAME_COPY_TO_CLIPBOARD_ASSET_WALLET = 'iframe-copy-to-clipboard-asset-wallet',
+  IFRAME_COPY_TO_CLIPBOARD_INTEGRITY_WALLET = 'iframe-copy-to-clipboard-integrity-wallet',
+  IFRAME_COPY_TO_CLIPBOARD_PRIVATE_KEY = 'iframe-copy-to-clipboard-private-key',
 }
```
sultanmyrza, line 10: **by ChatGPT** 
The code seems to be defining an enum for different post messages being sent from an iframe. There is nothing inherently wrong with this code, but some possible improvements could be:

1.   Adding comments to explain what each enum value means and when it should be used.
1.   Using more descriptive enum values instead of abbreviations to make it easier to understand the purpose of each value.
1.   Consolidating any duplicate enum values into a single value to reduce redundancy and improve clarity.

sultanmyrza: @tammyyang regarding ChatGPT suggestions:
1. to me it's completely selfexplanatory however for other devs or form me (after 2-3 month) it might make sense to have comments. 
2. opinionated.
3. I see no duplicates.

src/app/features/wallets/wallets.page.ts
tammyyang: @sultanmyrza for #3, will that lead to user confusion (because we silently skip it)?
For now, I think it is ok, let's see what users and QA find.

src/app/shared/iframe/iframe.ts
tammyyang: Agreed, they are selfexplanatory but a task to improve is even better

tammyyang (commented):
Review and commenting done.

src/app/features/wallets/wallets.page.ts
sultanmyrza: @tammyyang I think it's okay to skip postMessages. For example MetaMask plugin for chrome also broadcast some postMessages. When I listen for all postMessages I notice how MetaMask also broadcasts its own postMessages. 

So I believe skipping unknown postMessages silently make sense. And only react to capture app interested events

PR #2731, opened 2023-04-19
Feat v230321 capture app flavours should be built automatically by ci/cd

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1204430604757253) by [Unito](https://www.unito.io)

### DIFF
--- a/src/app/shared/apps-flyer/apps-flyer.service.ts
+++ b/src/app/shared/apps-flyer/apps-flyer.service.ts
@@ -24,7 +24,7 @@ export class AppsFlyerService {
 
   constructor(private readonly platform: Platform) {}
 
-  async initAppFlyerSDK() {
+  async initAppsFlyerSDK() {
     await this.platform.ready();
 
     if (this.shouldInit === false) return;
--- a/src/app/shared/apps-flyer/apps-flyer.service.ts
+++ b/src/app/shared/apps-flyer/apps-flyer.service.ts
@@ -26,13 +26,31 @@ export class AppsFlyerService {
 
   async initAppFlyerSDK() {
     await this.platform.ready();
+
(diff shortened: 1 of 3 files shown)


src/app/shared/apps-flyer/apps-flyer.service.ts
```diff
@@ -44,6 +62,6 @@ export class AppsFlyerService {
   }
 
   private get isNativePlatform() {
-    return this.platform.is('ios') || this.platform.is('android');
+    return this.platform.is('hybrid');
   }
 }
```
sultanmyrza, line 67: ChatGPT Plus code review with improvement suggestions:

1.  Rename the method `initAppFlyerSDK` to `initAppsFlyerSDK` for better consistency with the service name.
2.  Add a comment above the `shouldInit` getter to briefly explain its purpose.
3.  Remove the `if` statement in the `isNativePlatform` getter, as the `is('hybrid')` check should be sufficient to determine whether it's running on a native platform or not. You can directly return the result of the check:

```typescript

private get isNativePlatform() {
  return this.platform.is('hybrid');
}```

Overall, the code looks good, and the changes made in this commit address the SDK initialization logic. The logic for initializing the AppsFlyer SDK only if the dev key is provided and if it's a native platform is implemented correctly.

sultanmyrza: 1. TODO: Will add 1 commit to this PR (since it does not change any logic)
2. SKIP. Already selfdescriptive
3. SKIP. Extracting to `isNativePlatform` make more sense because `this.platform.is('hybrid')` doesn't tell much to a code reader/reviewer

shc261392 (approved):
Looks good to me

PR #2732, opened 2023-04-19
Feat v230321 set device information when user signing up

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1204430812544231) by [Unito](https://www.unito.io)

### DIFF
--- a/src/app/features/signup/signup.page.ts
+++ b/src/app/features/signup/signup.page.ts
@@ -6,7 +6,7 @@ import { TranslocoService } from '@ngneat/transloco';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { FormlyFieldConfig } from '@ngx-formly/core';
 import { combineLatest, defer } from 'rxjs';
-import { catchError, concatMapTo, first, tap } from 'rxjs/operators';
+import { catchError, concatMap, concatMapTo, first, tap } from 'rxjs/operators';
 import { BlockingActionService } from '../../shared/blocking-action/blocking-action.service';
 import { DiaBackendAuthService } from '../../shared/dia-backend/auth/dia-backend-auth.service';
 import { ErrorService } from '../../shared/error/error.service';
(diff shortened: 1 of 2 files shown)


src/app/features/signup/signup.page.ts
```diff
@@ -192,73 +192,79 @@ export class SignupPage {
...
+    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+    if (err instanceof HttpErrorResponse && err.status === 401)
+      return this.errorService.toastError$(
+        this.translocoService.translate('error.diaBackend.untrusted_client')
+      );
+    return this.errorService.toastError$(err);
+  }
 }
 
 interface SignupFormModel {
```
sultanmyrza, line 270: The code in this commit sends device information during the sign-up process. You've added a `concatMap` to chain the `device$` observable with the `createUser$` observable, which now includes the `fcm_token`, `platform`, and `device_identifier` in the request. The error handling has been moved to the `handleOnSubmitError` function to improve readability.

Here are a few suggestions for further improvements:

1.  Add comments to explain the purpose of the different observables, e.g., `device$`, `createUser$`, and `action$`.

2.  Change the `if` statements in the `handleOnSubmitError` function to a `switch` statement based on the error type, which can simplify the code and make it more readable.

```typescript

private handleOnSubmitError(err: unknown) {
  if (err instanceof HttpErrorResponse) {
    switch (err.error.error?.type) {
      case 'duplicate_email':
        return this.errorService.toastError$(
          this.translocoService.translate('error.diaBackend.duplicate_email')
        );
      case 'invalid_referral_code':
        return this.errorService.toastError$(
          this.translocoService.translate('error.diaBackend.invalid_referral_code')
        );
      // ...
    }

    if (err.error.error?.details?.username?.length > 0) {
      // ...
    }

    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
    if (err.status === 401) {
      // ...
    }
  }

  return this.errorService.toastError$(err);
}
```

Other than these suggestions, the changes made in this commit look good and effectively implement the feature to send device information during the sign-up process.

sultanmyrza: 1. SKIP:
2. SKIP: opinionated, if feels better.

src/app/shared/dia-backend/auth/dia-backend-auth.service.ts
```diff
@@ -550,3 +560,22 @@ type GetAvatarResponse = UploadAvatarResponse;
...
+  language: 'en-us' | 'zh-tw';
+  referral_code: string;
+  device?: UserDevice;
+}
+
+export interface UserDevice {
+  fcm_token: string;
+  platform: 'ios' | 'android' | 'web';
+  device_identifier: string;
+}
```
sultanmyrza, line 581: ChatGPT Plus code review with improvement suggestions:

1.  Add a comment to the `readDevice$()` function to explain its purpose and the values it returns. For example:

```typescrip
/**
 * Combines FCM token, device info, and device ID into an observable
 * to be used during the user creation process.
 *
 * @returns An observable that emits an array with FCM token, device info, and device ID.
 */
```

2.  In the `createUser$()` function, update the type of the `requestBody` variable from `Partial<SignUpRequestBody>` to `SignUpRequestBody` since all properties are required except for `device`. Remove the `any` type:

```typescript
const requestBody: SignUpRequestBody = {
  username,
  email,
  password,
  referral_code: referralCodeOptional,
  device,
};
```

3.  Remove the unnecessary `if (referralCodeOptional === '') delete requestBody.referral_code;` statement, as it's not needed with the `SignUpRequestBody` interface. The API will handle an empty `referral_code` correctly.

4.  For better readability, you can rename the `device` parameter in the `createUser$()` function to `userDevice` to differentiate it from the `Device` import:

```typescript
createUser$(
  username: string,
  email: string,
  password: string,
  referralCodeOptional: string,
  userDevice?: UserDevice
) {
```

5.  Add comments to the `UserDevice` and `SignUpRequestBody` interfaces to provide more context:

```typescript
/**
 * Represents a user's device information to be sent during the sign-up process.
 */
export interface UserDevice {
  fcm_token: string;
  platform: 'ios' | 'android' | 'web';
  device_identifier: string;
}

/**
 * Represents the request body for the sign-up API call.
 * See: https://dia-backend-dev.numbersprotocol.io/api/v3/redoc/#operation/auth_users_create
 */
export interface SignUpRequestBody {
  // ...
}
```

With these suggestions, your code should be more readable, and the overall structure should be clearer.

sultanmyrza: 1. SKIP: self explanatory
2. SKIP: irrelevant
3. SKIP: we need to check for '' empty string due to how SignUp page form works.
4. SKIP: opinionated
5. SKIP: Good suggestion but then we need to go all over the place and add comments for every <Somthing>RequestBody or developer can just check which endpoint it uses and go manually there. So I think adding comment is not super useful.

shc261392 (approved):
Looks good to me.

PR #2812, opened 2023-05-24
Fix v230413 issue error message when without device permission

sultanmyrza:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201016280880500/1204666164621170) by [Unito](https://www.unito.io)


src/app/features/home/custom-camera/custom-camera.page.ts
```diff
@@ -200,11 +205,53 @@ export class CustomCameraPage implements OnInit, OnDestroy {
...
+        'customCamera.error.canNotStartCamera'
+      );
+      this.errorService.toastError$(errMsg).subscribe();
+      return false;
+    }
+  }
+
   stopPreviewCamera() {
     this.customCameraService.stopPreviewCamera();
   }
```
sultanmyrza, line 257: **ChatGPT Plus code review with improvement suggestions:**

1\. **Exception Handling**: In the `canStartPreviewCamera` method, you catch a potential error but then don't provide much information about it. Even though you mention reporting it to Crashlytics, this isn't implemented yet. It would be useful to include more details in the error message or log the error in some way for debugging purposes. This could be done, for instance, by just console.logging the error for the time being.

2\. **Code Modularity**: Consider refactoring the permission checking and dialog presenting logic into separate methods. This would not only make the `canStartPreviewCamera` method shorter and easier to read, but also improve code reusability.

3\. **Promise Handling**: You're awaiting the `this.errorService.toastError$(errMsg).subscribe();` method call, which seems to return an Observable. The await keyword is used for Promises, not Observables. Consider whether this await is necessary or if you need to convert this Observable to a Promise.

4\. **Type Annotations**: In the catch block, you catch `_error: unknown`. While this is a valid type, it would be more useful to catch an Error type. This would give you access to the error message and other properties, and make the error easier to handle.

5\. **Method Complexity**: The method `canStartPreviewCamera` is doing several things: checking permissions, presenting a dialog, potentially navigating the user away, handling errors. According to the Single Responsibility Principle, it may be beneficial to separate some of these concerns into their own methods to improve readability and maintainability.

6\. **Terminology Consistency**: In your if statement you check if `permissions.camera !== 'granted'` and `permissions.microphone !== 'granted'`. While this is correct, it would be more intuitive to check if the permissions are equal to 'denied'. This way, the code reads more naturally and it's immediately clear what you're checking for.

Remember that these are general suggestions and your team or project might have specific coding standards or styles. Always ensure your improvements align with those requirements.

sultanmyrza: 1. Need to consult with Senior Devs wether we need to use firebase crashlytics.
2. Good point.
3. Still works but will change to `this.errorService.toastError$(errMsg).toPromise();` next time.
4. Will add typing if we will use _error varialbe later in firebase crashlytics.
5. Good point.
6. Opinionated

src/app/features/home/custom-camera/custom-camera.page.ts
```diff
@@ -200,11 +205,53 @@ export class CustomCameraPage implements OnInit, OnDestroy {
...
     await this.customCameraService.setCameraQuality(this.cameraQuality$.value);
     await this.syncCameraState();
   }
 
+  async canStartPreviewCamera() {
+    try {
+      const permissions = await this.customCameraService.requestPermissions();
+      if (
+        permissions.camera !== 'granted' ||
+        permissions.microphone !== 'granted'
```
shc261392, line 222: In current published app version do we always require the user to grant microphone permission in order to use camera? (even if user doesn't want to use video recording?)

sultanmyrza: @shc261392 to keep logic simple I ask all permissions at once. 

I will ask QA if we need lazy permission feature (aka ask microphone permission when user clicks record video only).

PR #2973, opened 2023-08-07
fix: automatically navigate back on iOS 16 and higher

sultanmyrza:
Fixes https://github.com/numbersprotocol/capture-lite/issues/2900


shc261392 (commented):
Review done 👍

PR #3237, opened 2024-03-13
feat(ActionService): add query param sort_field=index to sort network actions

shc261392:
## Changed
- Sort network actions by `index` field to allow control of display order

### DIFF
--- a/src/app/shared/actions/service/actions.service.ts
+++ b/src/app/shared/actions/service/actions.service.ts
@@ -14,7 +14,9 @@ export class ActionsService {
   getActions$() {
     return defer(() =>
       this.httpClient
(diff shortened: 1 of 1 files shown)


src/app/shared/actions/service/actions.service.ts
```diff
@@ -14,7 +14,9 @@ export class ActionsService {
   getActions$() {
     return defer(() =>
       this.httpClient
-        .get<GetActionsResponse<Action>>(`${BUBBLE_DB_URL}/api/1.1/obj/action`)
+        .get<GetActionsResponse<Action>>(
+          `${BUBBLE_DB_URL}/api/1.1/obj/action?sort_field=index`
```
olgahaha, line 18: suggest using `params` instead of appending it directly to the url

olgahaha (approved):

PR #3244, opened 2024-05-10
feat(features/home/details): download c2pa file

olgahaha:
### Added
1. Update the share menu in the asset detail page as follows: [#3244](https://github.com/numbersprotocol/capture-lite/pull/3244)
   - Add "Download C2PA" (new option)
   - Add "View Blockchain Proof" (old: View asset profile)
   - Update "Share Blockchain Proof" (current: Share Asset Profile)
   - Update "Copy Nid" (current: Copy IPFS address)

​
┆The issue is also created on [Asana](https://app.asana.com/0/1201016280880500/1207285567499306)


src/app/features/home/details/details.page.ts
```diff
@@ -174,6 +190,8 @@ export class DetailsPage {
     1
   );
 
+  private readonly shareMemuDismissed$ = new Subject<void>();
```
shc261392, line 193: Is it a typo? Should be `shareMenuDismissed$` instead?

src/app/features/home/details/details.page.ts
```diff
@@ -454,6 +524,65 @@ export class DetailsPage {
...
+      fileUrl = `file://${fileUrl}`;
+    }
+    await ShareService.shareFile(fileUrl)
+      .catch(async reason => {
+        if (reason?.message !== 'Share canceled') {
+          await this.errorService.toastError$(reason).toPromise();
+        }
+      })
+      .finally(() => Filesystem.deleteFile({ path: filePath }));
+  }
```
shc261392, line 584: Is it possible that on some sharing target platforms, shared files (especially image) are compressed and lost the C2PA injection data? As far as I know it happens at least on Slack and some major social media platforms.

Upload an C2PA image to slack -> download uploaded file -> get a file with different checksum and no C2PA claim

Would such situation happen when using the `shareFile` function? If yes we could probably consider sharing the IPFS url or add some warning about possible target platform compression.

olgahaha: There is a [strip](https://techcrunch.com/2020/05/11/slack-strips-location-data/) metadata mechanism on Slack. However, since the URL has just been added but not pinned to IPFS, I think it's better to download it immediately to prevent it from being garbage collected. Additionally, besides sharing on social media, the function can also be shared to app files or emails, so social media might be considered a limitation. I'll ask about it in the task, and we could update it in the future sprint if the UI or text should be modified.

shc261392 (approved):

PR #3255, opened 2024-10-25
feat(features/home): update profile name

olgahaha:
### Changed

1. Display and update profile name instead of username
1. Reduce duplicate API calls for profile

### Removed

1. Remove unused pages: `ProfilePage` and `PhoneVerificationPage`, and the service `DiaBackendAssetRefreshingService`

​
┆The issue is also created on [Asana](https://app.asana.com/0/1201016280880500/1208625719723109)


src/app/features/home/edit-profile/edit-profile.page.ts
```diff
@@ -27,22 +19,21 @@ import { NetworkService } from '../../../shared/network/network.service';
...
+  readonly avatar$ = this.diaBackendAuthService.profile$.pipe(
+    map(profile => profile.profile_picture_thumbnail)
+  );
+  readonly background$ = this.diaBackendAuthService.profile$.pipe(
     map(profile => profile.profile_background_thumbnail)
   );
   readonly form = new UntypedFormGroup({});
   model: EditProfileFormModel = {
-    username: '',
+    displayName: '',
```
shc261392, line 36: Found it in some cases it referred to as profileName (e.g. transaltion) and sometimes displayName. Maybe use profileName for all Ionic side variables and use `display_name` only on backend API response?

src/app/features/profile/phone-verification/phone-verification.page.html
shc261392, line 1: Just confirming, Is phone verification intended to be remove in the change?

olgahaha: I checked the code, and I didn’t find any references to it. I don’t think we need to update the code if there are no references that use the username. If I missed something, please correct me.

shc261392: Seems it's already removed prior to this change.

shc261392 (approved):

PR #3453, opened 2026-04-29
[Omni] Restrict Dependabot updates to production deps

numbers-official:
## Summary
This PR updates the Dependabot configuration to only process production dependencies. This reduces noisy Dependabot update failures caused by non-production dependency types while keeping the intended production dependency update flow intact.

## Changes
- Updated `.github/dependabot.yml` to apply `dependency-type: production` to Dependabot update rules.
- Limited Dependabot update scope so development-only dependencies are not included in the affected update checks.
- Preserved the existing Dependabot schedule and update behavior for production dependencies.
- Note that historical GitHub Actions error records will not be deleted, but new Dependabot runs should stop producing the same non-production dependency-type errors.

---
Generated by Olga Shen with [Omni](https://omniai.one/)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,11 @@ updates:
       interval: 'weekly'
     target-branch: 'main'
     open-pull-requests-limit: 10
+    # Only automate production dependency updates. Development/build toolchain
+    # advisories are reviewed manually because they often require coordinated
+    # framework upgrades or upstream fixes.
+    allow:
+      - dependency-type: 'production'
     # Skip major-version PRs. Majors for Angular / ngrx / ionic / capacitor etc.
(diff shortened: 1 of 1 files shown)


numbers-official:
Reviewed for regression risk. No blocking issues found.\n\nScope checked:\n- Change is limited to .github/dependabot.yml.\n- GitHub Dependabot docs list allow.dependency-type: production as supported for npm, and allow is applied before ignore, so the existing semver-major ignore remains effective.\n- Parsed the updated YAML locally with PyYAML: version=2, one npm update block for '/', allow=[{dependency-type: production}], existing ignore/groups preserved.\n- Ran git diff --check for the changed file; no whitespace issues.\n\nI did not run full app lint/build because this PR only changes Dependabot config and node_modules is not installed in the review workspace.

